PR #1, opened 2022-08-01
Jt/dev

ckousik:
Opened this PR for review


src/listener.ts
```diff
@@ -0,0 +1,68 @@
+import { WebRTCListenerOptions } from './options.js'
```
ckousik, line 1: This file is unnecessary as we do not support listening on the browser

src/socket-to-conn.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export interface ToMultiaddrConnectionOptions extends AbortOptions {
+  remoteAddr: Multiaddr
+}
+
+/**
+ * Convert a socket into a MultiaddrConnection
+ * https://github.com/libp2p/js-libp2p-interfaces/tree/master/src/transport#multiaddrconnection
+ */
+export function toMultiaddrConnection (socket: WebRTCPeer, options: ToMultiaddrConnectionOptions): MultiaddrConnection {
```
ckousik, line 18: I don't think we want to expose a `MultiaddrConnection`. This will need to use the default upgrader which would cause nested muxing and encryption (noise over DTLS). This will make this implementation incompatible with the spec and the Rust and Go implementations. We should instead directly implement the `Connection` interface.

src/socket.ts
```diff
@@ -0,0 +1,54 @@
+import type { Socket } from 'socket.io-client'
```
ckousik, line 1: Not sure if we want to use the `webrtc-peer` library. It should be simpler to just use the RTCPeerConnection.

src/transport.ts
```diff
@@ -1,16 +1,37 @@
...
 	async dial(ma: Multiaddr, options: DialOptions): Promise<Connection> {
+		const rawConn = this._connect(ma, options)
+		const maConn = toMultiaddrConnection(rawConn, { remoteAddr: ma, signal: options.signal })
+		log('new outbound connection %s', maConn.remoteAddr)
 		throw new Error("not implemented");
 	}
 
 	createListener(options: CreateListenerOptions): Listener {
-		throw new Error("not implemented");
+		return createListener(options.handler ?? noop, this.components.getPeerId(), this, options)
```
ckousik, line 34: this should throw an error

src/transport.ts
```diff
@@ -1,16 +1,37 @@
...
+
+function ma2sdp(ma: Multiaddr): RTCSessionDescription {
+	return new RTCSessionDescription(JSON.parse("{TODO: \"" + ma + "\"}"));
+}
 
 export class WebRTCTransport implements Transport {
 
 	async dial(ma: Multiaddr, options: DialOptions): Promise<Connection> {
+		const rawConn = this._connect(ma, options)
+		const maConn = toMultiaddrConnection(rawConn, { remoteAddr: ma, signal: options.signal })
```
ckousik, line 28: we should be implementing the connection interface instead of `MultiaddrConnection`.

src/transport.ts
```diff
@@ -24,4 +45,23 @@ export class WebRTCTransport implements Transport {
 	get [symbol](): true {
 		return true
 	}
+
+	todo_cb() {
+	}
+
+	_connect (ma: Multiaddr, options: WebRTCDialOptions) {
+		this.channel = this.peer_connection.createDataChannel("channel");
```
ckousik, line 53: This isn't part of the spec yet, but we have been using the label `"data"` here. Also, this channel should have `{ negotiated: true, id: 1 }` in the creation options, as this is negotiated as part of the SDP.

src/transport.ts
```diff
@@ -24,4 +45,23 @@ export class WebRTCTransport implements Transport {
...
+		this.channel.onopen = this.todo_cb;
+		this.channel.onclose = this.todo_cb;
+		this.peer_connection.onicecandidate = this.todo_cb;
```
ckousik, line 56: This can be a noop as we have no direct way of exchanging candidates with the remote. The exchange happens via the SDP which is constructed by the server after looking at incoming STUN messages.

src/transport.ts
```diff
@@ -24,4 +45,23 @@ export class WebRTCTransport implements Transport {
...
+		this.peer_connection.setRemoteDescription(ma2sdp(ma));
```
ckousik, line 57: Refer https://github.com/libp2p/go-libp2p/pull/1655/files#diff-4f644a34f6d6d3f5ce6ddc58c2a2ff7e5dc3312006ec02810ef33a8016c529a1R52 for SDP construction.

src/transport.ts
```diff
@@ -24,4 +45,23 @@ export class WebRTCTransport implements Transport {
...
+		this.peer_connection.createOffer()
```
ckousik, line 58: This should happen immediately after creating the data channel.

src/transport.ts
```diff
@@ -24,4 +37,21 @@ export class WebRTCTransport implements Transport {
 	get [symbol](): true {
 		return true
 	}
+
+	todo_cb() {
+	}
+
+	_connect (ma: Multiaddr, options: WebRTCDialOptions) {
+		this.channel = this.peer_connection.createDataChannel("data");
```
ckousik, line 45: `this.peer_connection` would be incorrect. A peerconnection should be scoped to a connection and not a transport.

PR #2, opened 2022-08-02
WIP

John-LittleBearLabs:
## TODO

- [X] Skeletons for Transport, Connection, and Stream implementations
- [ ] SDP munging
- [X] SDP creation
- [ ] Implement Transport.dial
- [ ] Implement remainder of Transport
- [ ] Implement Connection.newStream
- [ ] Implement remainder of Connection
- [ ] Implement Stream.source
- [ ] Implement Stream.sink
- [ ] Implement remainder of Stream
- [ ] Fill out README.md and any other worthwhile documentation
- [ ] Unit test SDP utilities
- [ ] Unit test Transport
- [ ] Unit test Connection
- [ ] Unit test Stream
- [ ] Compliance testing
- [ ] End-to-end test


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
ddimaria, line 1: You may want to remove this file (and .gitignore it).

package.json
```diff
@@ -0,0 +1,27 @@
...
+    "aegir": "^37.4.6"
+  },
+  "dependencies": {
+    "@libp2p/components": "^2.0.1",
+    "@libp2p/interface-transport": "^1.0.2",
+    "@libp2p/interfaces": "^3.0.3",
+    "@libp2p/logger": "^2.0.0",
+    "abortable-iterator": "^4.0.2",
+    "socket.io-client": "^4.1.2",
+    "typescript": "^4.7.4",
```
ddimaria, line 24: Dev dependency?

src/connection.ts
```diff
@@ -0,0 +1,66 @@
+import { Connection }  from '@libp2p/interface-connection'
+import {ConnectionStat}from '@libp2p/interface-connection'
```
ddimaria, line 2: Some inconsistent formatting.  Try installing `prettier` and add to `format on save` in you editor.

src/connection.ts
```diff
@@ -0,0 +1,66 @@
...
+    remoteAddr: Multiaddr
+    direction: Direction
+    tags?: string[]
+    stat: ConnectionStat
+    pc:  RTCPeerConnection
+    credential_string: string
+}
+
+export class WebRTCConnection implements Connection {
+    id: string;
```
ddimaria, line 26: Consider adding access modifiers here.

src/connection.ts
```diff
@@ -0,0 +1,66 @@
...
+        throw new Error("not implemented")
+    }
+    removeStream(id: string): void {
+        throw new Error("not implemented")
+    }
+    async close(): Promise<void> {
+        throw new Error("not implemented")
+    }
+
+    private peerConnection: RTCPeerConnection;
```
ddimaria, line 62: You may want to move these up to the top of the class with the other class members.

src/stream.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Stats about this stream
+   */
+  stat: StreamStat;
+
+  /**
+   * User defined stream metadata
+   */
+  metadata: Record<string, any>;
```
ddimaria, line 63: Do we know more about the type where `any` is now?

src/transport.ts
```diff
@@ -0,0 +1,46 @@
...
+import {Listener, Transport } from '@libp2p/interface-transport'
+import {DialOptions, symbol } from '@libp2p/interface-transport'
+import { logger }             from '@libp2p/logger'
+import { Multiaddr }          from '@multiformats/multiaddr';
+import { v4 as genUuid }      from 'uuid';
+
+const log = logger('libp2p:webrtc:transport')
+
+export class WebRTCTransport implements Transport {
+
```
ddimaria, line 14: Mix of tabs and spaces in diff files.  We should find out Protocol's preferred tab/space.

src/transport.ts
```diff
@@ -0,0 +1,46 @@
...
+		log('new outbound connection %s', rawConn, genUuid());
+		throw new Error("not implemented");
+	}
+
+	createListener(options: CreateListenerOptions): Listener {
+		throw new Error("TODO - replace with an exception more appropriate to the fact that this will not be implemented.");
+	}
+
+	filter(multiaddrs: Multiaddr[]): Multiaddr[] {
+		return []
```
ddimaria, line 26: Add semicolon here and below for consistency.

test/node.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-env mocha */
```
ddimaria, line 1: Consider a different name for this file.

src/connection.ts
John-LittleBearLabs: The aim there was aligning `from`. 

Changed my settings as recommended, though. Should be prettier in the future.

src/connection.ts
John-LittleBearLabs: What is the proper convention for data members that are satisfying some interface?

No modifier was present in the code I stole this from. Perhaps I should specify `public` ?

src/connection.ts
John-LittleBearLabs: Sure thing - whatever the convention is.

I would think I'd want to separate members that exist to satisfy the interface from those that exist for implementation's sake.

src/connection.ts
ddimaria: I don't think there is a convention for separating out data members, though I've seen the sorted by access modification type in some repos.

You can be explicit about `public` or leave it off, it's stylistic.  If they're not meant to be used outside of this class, consider making all `private` until they need to be `public`.

src/stream.ts
John-LittleBearLabs: At this point I don't think it's been settled. Right now it's just directly copied out of the Stream interface.

src/connection.ts
ddimaria: I don't think I've ever seen that done in TS.

src/stream.ts
ddimaria: Ah, I guess it's OK there then.

src/connection.ts
John-LittleBearLabs: I think they can't be private. And given that, sorting by access modifier accomplishes what I was after.

````
src/connection.ts:25:14 - error TS2420: Class 'WebRTCConnection' incorrectly implements interface 'Connection'.
  Property 'id' is private in type 'WebRTCConnection' but not in type 'Connection'.

25 export class WebRTCConnection implements Connection {
````

src/transport.ts
John-LittleBearLabs: I could probably borrow this:

https://github.com/protocol/arg/blob/master/.prettierrc

test/node.ts
John-LittleBearLabs: The name seems to be the norm for `aegir test`, which I'm borrowing from *star.

If I _just_ rename it without doing stuff to the package.json I get

````
Error: No test files found
Command failed with exit code 1: mocha test/node.{js,cjs,mjs} test/**/*.spec.{js,cjs,mjs} dist/test/node.{js,cjs,mjs} dist/test/**/*.spec.{js,cjs,mjs} --ui bdd --require source-map-support/register --timeout=60000 --bail
error Command failed with exit code 1.
````

I think I can make it work, though.

````
$./node_modules/.bin/mocha dist/test/notnode.js


  0 passing (2ms)
````

Do you have a preferred name? tests.ts ?

ddimaria: If there will only be a single file in this folder, maybe `index.ts`?  If you'll mix unit and e2e, then `unit.ts`.

John-LittleBearLabs:
Moved to https://github.com/little-bear-labs/js-libp2p-webrtc/pull/4 during branch rename.

PR #5, opened 2022-08-05
intitial stream logic

ckousik:
Implemented a simple stream API for the DataChannel.

### DIFF
--- a/src/stream.ts
+++ b/src/stream.ts
@@ -1,53 +1,21 @@
-import { Stream } from '@libp2p/interface-connection';
+import { Stream, Direction } from '@libp2p/interface-connection';
 import { StreamStat } from '@libp2p/interface-connection';
-import { logger } from '@libp2p/logger';
+// import { logger } from '@libp2p/logger';
 import { Source } from 'it-stream-types';
 import { Sink } from 'it-stream-types';
+import { pushable, Pushable } from 'it-pushable';
+import defer, { DeferredPromise } from 'p-defer';
+import merge from 'it-merge';
 
-const log = logger('libp2p:webrtc:connection');
+// const log = logger('libp2p:webrtc:connection');
+
+type StreamInitOpts = {
+  channel: RTCDataChannel;
+  direction: Direction;
+  metadata?: Record<string, any>;
+};
 
 export class WebRTCStream implements Stream {
-  constructor() {
-    this.id = 'TODO';
-    this.stat = {
-      direction: 'outbound',
-      timeline: {
-        open: 0,
-        close: 0,
-      },
-    };
-    this.metadata = {};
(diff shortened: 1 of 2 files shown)


src/stream.ts
```diff
@@ -1,70 +1,140 @@
-import { Stream } from '@libp2p/interface-connection';
+import { Stream, Direction } from '@libp2p/interface-connection';
 import { StreamStat } from '@libp2p/interface-connection';
-import { logger } from '@libp2p/logger';
+// import { logger } from '@libp2p/logger';
```
John-LittleBearLabs, line 3: You don't need/want any logging?
I suppose that makes sense. The Stream's methods could be called very frequently.

src/stream.ts
```diff
@@ -1,70 +1,140 @@
...
   /**
    * Close a stream for reading and writing
    */
-  close(): void {}
+  close(): void {
+    if (this.closed) {
+      return;
+    }
+    this.closed = true;
+    this.closeRead();
```
John-LittleBearLabs, line 106: It looks like there will never be a time where either
`readClosed && !closed`
or
`writeClosed && !closed`
So I would think those if statements that check for `closed || xClosed` could simply be `closed`. No?

John-LittleBearLabs: Oh... I see closeRead() and closeWrite() are public and in the interface. So my above assumption is wrong. Sorry.

src/stream.ts
```diff
@@ -1,70 +1,140 @@
...
-   * User defined stream metadata
-   */
-  metadata: Record<string, any>;
-
-  source: Source<Uint8Array> = process.stdin; //TODO
-  sink: Sink<Uint8Array, Promise<void>>;
-
-  private dataChannel?: RTCDataChannel;
+  reset(): void {
+    this.close();
```
John-LittleBearLabs, line 138: Should this, and also some other methods, alter `this.stat`?

src/stream.ts
```diff
@@ -1,70 +1,140 @@
...
+      }
+      (this.source as Pushable<Uint8Array>).push(evt.data);
+    };
+    this.channel.onclose = (_) => this.close();
+    this.channel.onerror = (_event) => {
+      this.abort(new Error('TODO'));
+    };
   }
 
+  // duplex sink
```
John-LittleBearLabs, line 96: Is this comment where it belongs?

John-LittleBearLabs (approved):

src/stream.ts
ckousik: I got the logic operators wrong. My mistake.

src/stream.ts
ckousik: Yes, I will add that in a future commit.

John-LittleBearLabs (approved):
LGTME

PR #11, opened 2022-08-10
Transport implements initializable

ckousik:
@John-LittleBearLabs Could we merge this into your existing PR? I can then work on connection using your branch. It allows access to the registrar and metrics components.


src/transport.ts
```diff
@@ -41,6 +45,7 @@ export class WebRTCTransport implements Transport {
   }
 
   async _connect(ma: Multiaddr, options: WebRTCDialOptions) {
+    let registrar = (await this.components.promise).getRegistrar();
```
John-LittleBearLabs, line 48: What is registrar used for?

PR #14, opened 2022-08-11
[CON-431][CON-430][CON-438] Implement Connection

ckousik:
Implement connection interface for RTCPeerConnection.


package.json
```diff
@@ -9,28 +9,35 @@
   "scripts": {
     "build": "aegir build",
     "clean": "rm -rfv node_modules dist *.lock *-lock.json ",
-    "test": "aegir test",
+    "test": "aegir test --target browser",
```
John-LittleBearLabs, line 12: The example suggested targets for test, test:browser, and test:node.
But I suppose this project only targets a browser, right, so maybe we should only do that?

src/connection.ts
```diff
@@ -4,70 +4,235 @@ import { AbortOptions } from '@libp2p/interfaces';
...
+    this.remotePeer = init.remotePeer;
+    this.localPeer = init.localPeer;
+    this.localAddr = init.localAddr;
+    this.components = init.components;
     this.stat = {
-      direction: 'outbound',
-      timeline: { open: 0 },
-      status: 'CLOSED',
+      direction: init.direction,
+      status: 'OPEN',
```
John-LittleBearLabs, line 56: Is it always open immediately?

src/connection.ts
```diff
@@ -4,70 +4,235 @@ import { AbortOptions } from '@libp2p/interfaces';
...
-  async newStream(multicodecs: string | string[], options?: AbortOptions): Promise<ic.Stream> {
-    // let label = uuid.v4()
-    // let dc = this._peerConnection.createDataChannel(label, {})
-    // await datachannel opening
-    // return DataChannelStream(dc)
-    log('TODO', this.ufrag);
-    this.peerConnection.createDataChannel(genUuid());
-    throw new Error('not implemented');
+  private wrapMsStream(rawStream: WebRTCStream, stream: Duplex<Uint8ArrayList, Uint8ArrayList | Uint8Array, Promise<void>>): ic.Stream {
+    return {
```
John-LittleBearLabs, line 113: Is this wrapping just to be able to set the ID? Because if so... we could add that as an optional thing to StreamInit

src/connection.ts
```diff
@@ -4,70 +4,235 @@ import { AbortOptions } from '@libp2p/interfaces';
...
+
+    this.addStream(result);
+    return result;
   }
 
   addStream(stream: ic.Stream): void {
-    throw new Error('not implemented');
+    let protocol = stream.stat.protocol!;
+    let direction = stream.stat.direction;
+    if (this.countStream(protocol, direction) === this.findStreamLimit(protocol, direction)) {
```
John-LittleBearLabs, line 218: Although I  understand it would be an incredibly bizarre implementation that did this, it seems to me to be at least _possible_ that the data being fetched by findStreamLimit _could_ theoretically change at runtime. So I'd be inclined to test with something like `>=`

src/connection.ts
```diff
@@ -4,70 +4,235 @@ import { AbortOptions } from '@libp2p/interfaces';
...
+
+  async newStream(protocols: string | string[], options: AbortOptions = {}): Promise<ic.Stream> {
+    let label = genUuid().slice(0, 8);
+    let openPromise = defer();
+    let abortedPromise = defer();
+    let controller: TimeoutController | undefined;
+    let metrics = this.components.getMetrics();
+    let openError: Error | undefined;
+
+    log.trace(`opening new stream with protocols: ${protocols}`);
```
John-LittleBearLabs, line 155: I see that we're not testing stream limits here. So those limits _only_ apply to incoming streams, not the sum of the two?

src/transport.ts
```diff
@@ -10,22 +10,24 @@ import { CreateListenerOptions, DialOptions, Listener, symbol, Transport } from
...
+    this.componentsPromise.resolve();
+    this.components = components;
   }
 
   async dial(ma: Multiaddr, options: DialOptions): Promise<Connection> {
-    const rawConn = this._connect(ma, options);
-    log('new outbound connection %s', rawConn, genUuid());
-    throw new Error('not implemented');
+    const rawConn = await this._connect(ma, options);
+    log(`dialing address - ${ma}`);
```
John-LittleBearLabs, line 29: Given this happens after `_connect`, shouldn't it be "dialed" address?

test/util.ts
```diff
@@ -0,0 +1,110 @@
...
+import {v4} from 'uuid';
+import {Registrar, StreamHandler} from '@libp2p/interface-registrar';
+import { pipe } from 'it-pipe';
+import { logger } from '@libp2p/logger';
+
+const log = logger('libp2p:webrtc:test:util');
+
+export const echoHandler: StreamHandler = ({ stream }) => pipe(stream.source, stream.sink);
+
+export async function createConnectedRTCPeerConnectionPair(): Promise<RTCPeerConnection[]> {
```
John-LittleBearLabs, line 17: :+1:

John-LittleBearLabs (approved):

package.json
ckousik: Yes, this is because we don't intend to support node at the moment.

PR #16, opened 2022-08-12
Getting more specific with the errors being thrown.

John-LittleBearLabs:


ckousik:
`js-libp2p` has error codes which we should try to integrate here

John-LittleBearLabs:
> `js-libp2p` has error codes which we should try to integrate here

@ckousik It looks like they're using
https://www.npmjs.com/package/err-code
Do you mean that we should use that, too, or that we should try to adopt their specific codes?
https://github.com/libp2p/js-libp2p/blob/master/src/errors.ts
I'm not sure that there's an easy way to do code sharing here, beyond copypasta.

John-LittleBearLabs:
> `js-libp2p` has error codes which we should try to integrate here

@ckousik Updated to use err-code.

ckousik:
@John-LittleBearLabs We can also import the `js-libp2p` package and use their error codes, but this should suffice for our use case.

PR #19, opened 2022-08-16
Unit testing WebRTCTransport

John-LittleBearLabs:
Unit tests for Transport class

Also an impl for filter

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,7 @@
   "type": "module",
   "scripts": {
     "build": "aegir build",
-    "test": "aegir test",
-    "test:browser": "aegir test --target browser",
+    "test": "aegir test --target browser",
     "format": "prettier --write src/*.ts"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   "type": "module",
   "scripts": {
     "build": "aegir build",
-    "clean": "aegir clean ; rm -rf node_modules/ ; rm -f package-lock.json yarn.lock",
     "test": "aegir test --target browser",
     "format": "prettier --write src/*.ts"
   },
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "type": "module",
   "scripts": {
     "build": "aegir build",
-    "clean": "aegir clean ; rm -rf node_modules/ ; rm package-lock.json ; rm yarn.lock",
+    "clean": "aegir clean ; rm -rf node_modules/ ; rm -f package-lock.json yarn.lock",
     "test": "aegir test --target browser",
     "format": "prettier --write src/*.ts"
   },
--- a/src/sdp.ts
+++ b/src/sdp.ts
@@ -14,23 +14,6 @@ const mbdecoder = (function () {
 })();
 
 const CERTHASH_CODE: number = 466;
-const ANSWER_SDP_FORMAT: string = `v=0
-o=- 0 0 IN {IPVERSION} {IP}
-s=-
-c=IN {IPVERSION} {IP}
-t=0 0
-a=ice-lite
-m=application {PORT} UDP/DTLS/SCTP webrtc-datachannel
-a=mid:0
-a=setup:active
-a=ice-options:ice2
-a=ice-ufrag:{UFRAG}
-a=ice-pwd:{UFRAG}
-a=fingerprint:{CERTFP}
-a=sctp-port:5000
-a=max-message-size:100000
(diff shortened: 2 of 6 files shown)


src/transport.ts
```diff
@@ -39,7 +39,7 @@ export class WebRTCTransport implements Transport, Initializable {
   }
 
   filter(multiaddrs: Multiaddr[]): Multiaddr[] {
-    return [];
+    return multiaddrs.filter(validMa);
```
John-LittleBearLabs, line 42: @ckousik note that this is hiding in here.

test/transport.browser.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    
+    it('symbol property getter', () => {
+        let t = new underTest.WebRTCTransport();
+        let s = t[symbol];
+        expect(s).to.equal(true);
+    });
+
+    it('filter gets rid of some invalids and returns a valid', async () => {
+        let mas: Multiaddr[] = [
+            '/ip4/1.2.3.4/udp/1234/webrtc/certhash/uEiAUqV7kzvM1wI5DYDc1RbcekYVmXli_Qprlw3IkiEg6tQ',
```
John-LittleBearLabs, line 58: It may be of interest to note, I did attempt to test filter in the case where a certhash is present but missing a character.
But filter takes as an argument a Multiaddr, not a string. And the constructor will throw an Error on that input, so you don't even get to call filter in the first place.

package.json
```diff
@@ -8,8 +8,8 @@
   "type": "module",
   "scripts": {
     "build": "aegir build",
-    "test": "aegir test",
-    "test:browser": "aegir test --target browser",
+    "clean": "aegir clean ; rm -rf node_modules/ ; rm -f package-lock.json yarn.lock",
```
ddimaria, line 11: Curious as to why there's a need to remove the lock files?

John-LittleBearLabs: Dumb paranoia on my part. I'm fine with getting rid of that script - don't think anyone else is running it anyhow.

src/sdp.ts
```diff
@@ -85,13 +86,12 @@ function certhashToFingerprint(ma: Multiaddr): string {
...
 function ma2sdp(ma: Multiaddr, ufrag: string): string {
   return ANSWER_SDP_FORMAT.replace('%s', ipv(ma))
-    .replace('%s', ip(ma))
-    .replace('%s', ipv(ma))
-    .replace('%s', ip(ma))
-    .replace('%d', port(ma).toString())
-    .replace('%s', ufrag)
-    .replace('%s', ufrag)
-    .replace('%s', certhashToFingerprint(ma));
+    .replaceAll('{IP}', ip(ma))
```
ddimaria, line 89: You could avoid this replacing by using builtin string interpolation and/or tagged templates: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/transport.ts
```diff
@@ -74,25 +80,32 @@ export class WebRTCTransport implements Transport, Initializable {
...
     handshakeDataChannel.onopen = (_) => dataChannelOpenPromise.resolve();
-    setTimeout(dataChannelOpenPromise.reject, 10000);
+    handshakeDataChannel.onerror = (ev: Event) => {
+      log.error('Error opening a data channel for handshaking: %s', ev.toString()); 
+        dataChannelOpenPromise.reject();
+      };
+    setTimeout(() => {
+        log.error('Data channel never opened. State was: %s', handshakeDataChannel.readyState.toString()); 
+        dataChannelOpenPromise.reject() 
+      }, 10000);
```
ddimaria, line 103: Possible magic number.  Can you assign it to a variable?  Also, can you comment on the significance of the number chose?

test/sdp.spec.ts
```diff
@@ -21,11 +23,33 @@ a=max-message-size:100000
 
 describe('SDP creation', () => {
   it('handles simple blue sky easily enough', async () => {
+    return;
```
ddimaria, line 26: Nice tests 💯

ddimaria (approved):

ddimaria:
@John-LittleBearLabs just one little comment on a magic number, can merge after that.

src/transport.ts
John-LittleBearLabs: It's a completely arbitrary number. I believe Chinmay used it when he was demonstrating the technique, and I stole it. 

10 seconds does work as a timeout in that it's considerably longer than this process should ever take. But also, yes, it would be completely reasonable to name this variable.

PR #20, opened 2022-08-26
Add message framing to support half-close and reset of stream

John-LittleBearLabs:
https://github.com/mxinden/specs/pull/1


mxinden:
Great to see this.

I am curious what you think of the proposed framing mechanism. We need it for the browser side, thus we should adjust it to the needs of this JS implementation.

John-LittleBearLabs:
As an aside, this diff will become cleaner when https://github.com/little-bear-labs/js-libp2p-webrtc/pull/19 is merged into develop, since there are changes in common.

@mxinden I haven't yet run into anything with this mechanism that is specially/unusually difficult to implement on the browser side. I had misunderstood the discussion of the varint length prefix at first, so perhaps we could at some point attempt to add some clarity for newbies' sake.

But then again, I clearly haven't finished. In particular I haven't attempted to handle flags being sent to the browser, which conceptually should be roughly half the proposal.

src/stream.ts
```diff
@@ -179,7 +205,25 @@ export class WebRTCStream implements Stream {
...
     this.stat = defaultStat(this.stat.direction);
+    this.writeClosed = true;
+    this.closeWritePromise.resolve();
+    if (this.readClosed && this.writeClosed) {
+      this.close();
+    }
+  }
+
+  private _sendFlag(flag: pb.Message_Flag): void {
+    if (this.writeClosed) {
```
ckousik, line 218: we should still be able to send a flag if write is closed. This is useful in case of stream resets.

package.json
```diff
@@ -7,6 +7,8 @@
   "license": "Apache-2.0 or MIT",
   "type": "module",
   "scripts": {
+    "autogen": "npx protoc --ts_out proto_ts --proto_path src src/*.proto",
```
ddimaria, line 10: Can you document these in the README?

src/message.proto
```diff
@@ -0,0 +1,20 @@
...
+    FIN = 0;
+    // The sender will no longer read messages on the stream. Incoming data is
+    // being discarded on receipt.
+    STOP_SENDING = 1;
+    // The sender abruptly terminates the sending part of the stream. The
+    // receiver can discard any data that it already received on that stream.
+    RESET = 2;
+  }
+
+  optional Flag flag=1;
```
ddimaria, line 17: nit: can you make the spacing consistent with the below `flag = 1;`

src/stream.ts
```diff
@@ -97,7 +98,27 @@ export class WebRTCStream implements Stream {
...
+          log.trace('Remote has indicated, with "STOP_SENDING" flag, that it will discard any messages we send.');
+          this.closeWrite();
+          break;
+        case pb.Message_Flag.FIN:
+          log.trace('Remote has indicated, with "FIN" flag, that it will not send any further messages.');
+          this.closeRead();
+          break;
+        case pb.Message_Flag.RESET:
+          log.trace('Remote abruptly stopped sending, indicated with "RESET" flag.');
+          this.closeRead();
```
ddimaria, line 116: Is there the possibility that this isn't exhaustive and needs a default case?

test/connection.browser.spec.ts
```diff
@@ -3,47 +3,46 @@
 import {createConnectionPair, echoHandler} from "../test/util.js";
 import { expect } from 'aegir/chai';
 import { pipe } from 'it-pipe';
-import all from 'it-all';
 import first from 'it-first';
 import {fromString} from 'uint8arrays/from-string';
 import {v4} from 'uuid';
+// import { enable as enableLogger } from '@libp2p/logger';
```
ddimaria, line 9: remove?

test/connection.browser.spec.ts
```diff
@@ -3,47 +3,46 @@
...
+    clientStream.sink([data]);
+    let responsed = false;
+    for await (const response of clientStream.source) {
+      expect(response).to.not.be.undefined;
+      expect(response.subarray()).to.equalBytes(data);
+      responsed = true;
+      break;
+    }
+    expect(responsed).to.be.true();
+  }).timeout(54321);
```
ddimaria, line 45: Curious what the significance of this number is?

ddimaria (approved):
Just added some non-critical comments/questions.

src/stream.ts
John-LittleBearLabs: What would a default case do? Throw an exception?

test/connection.browser.spec.ts
John-LittleBearLabs: It's larger than the default timeout. It may be a vestige of a time when I was trying to figure out why the connection is timing out. Will try removing it.

John-LittleBearLabs: Yeah, works fine.

src/stream.ts
ddimaria: I don't know enough of the context to know, but I guess the `undefined` case is acting like the default case.  I was just thinking of the possibility of other values not here, like `null` or whatever else is possible.  I think you can ignore this.

John-LittleBearLabs: `undefined` is there because the flag is optional in the protobuf. If it's missing it means it's a normal packet doing normal streaming things for the application (the protocol allows for the flag and message to both be present, but in practice it's one or the other).

In Javascript if it's a value that's not handled and there's no default, does it skip the whole switch like it would in C ? If so I think that's very reasonable behavior if someone tries to throw null at us somehow (presumably a coding error).

mxinden: Following up here, I think the implementation should support all possible cases:

- no data and no flag: Just ignore the hole packet. Could e.g. be possible in case we extend the Protobuf schema in a backwards compatible way in the future.
- data and no flag
- data and flag
- no data and flag

John-LittleBearLabs: @mxinden The `switch` and `if` seem to be independent to me, so the behavior you're describing is what I believe to already be the case.

For example, the `FIN` `case` in `switch` does a `break` not a `return`, so after handling that you head on to the `if` check for `message`

But if you got undefined/undefined you'd break out of the switch without doing anything and then pass over the if.

PR #23, opened 2022-09-12
Interop test between Go and Js implementations

ckousik:
This PR creates an interoperability test between the Go and Js implementation and aims to automate the test in CI.


src/connection.ts
```diff
@@ -91,21 +91,25 @@ export class WebRTCConnection implements ic.Connection {
 
       log.trace(`${logPrefix} supported protocols - ${protocols}`);
```
GlenDC, line 92: out of scope for this PR but why do we not inject the formatting logic (e.g. `logPrefix`) as part of our logging setup/functions? Seems a bit silly that we need to add it for each line here.

src/connection.ts
```diff
@@ -91,21 +91,25 @@ export class WebRTCConnection implements ic.Connection {
...
+        this.addStream(result);
 
-      // handle stream
-      let { handler } = registrar.getHandler(protocol);
-      handler({ connection: this, stream: result });
+        // handle stream
+        let { handler } = registrar.getHandler(protocol);
+        handler({ connection: this, stream: result });
+      } catch (err) {
+        log.error('stream error: ', rawStream.id, rawStream.stat.direction);
```
GlenDC, line 111: Do we really want to silence an error from this connection function? Will this not make it harder to use, as the user won't really be able to handle unexpected errors in this method cal?

src/options.ts
```diff
@@ -6,5 +6,4 @@ export interface WebRTCListenerOptions extends CreateListenerOptions {
   //   channelOptions?: WebRTCReceiverInit
 }
 
-export interface WebRTCDialOptions extends DialOptions {
-}
+export interface WebRTCDialOptions extends DialOptions {}
```
GlenDC, line 9: our linting rules must be pretty relaxed as normally you would get an "empty interface" complaint here, no?

src/sdp.ts
```diff
@@ -62,16 +62,16 @@ function certhashToFingerprint(ma: Multiaddr): string {
   }
 
   let fp = mhdecoded.digest.reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '');
```
GlenDC, line 64:
```suggestion
  const fp = mhdecoded.digest.reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '');
```

again this shows me that our linting is pretty relaxed, as I would expect the linter to complain about this kind of stuff?

src/sdp.ts
```diff
@@ -62,16 +62,16 @@ function certhashToFingerprint(ma: Multiaddr): string {
...
-  fp = fp.match(/.{1,2}/g)!.join(':');
+  let fpSdp = fp.match(/.{1,2}/g)!.join(':');
```
GlenDC, line 65:
```suggestion
  const fpSdp = fp.match(/.{1,2}/g)!.join(':');
```

src/sdp.ts
```diff
@@ -80,14 +80,13 @@ t=0 0
...
-a=setup:active
-a=ice-options:ice2
+a=setup:passive
 a=ice-ufrag:${ufrag}
 a=ice-pwd:${ufrag}
 a=fingerprint:${CERTFP}
 a=sctp-port:5000
 a=max-message-size:100000
-a=candidate:1 1 UDP 1 ${IP} ${PORT} typ host`;
+a=candidate:1467250027 1 UDP 1467250027 ${IP} ${PORT} typ host\r\n`;
```
GlenDC, line 89: where do these numbers come from, might be nice to add a line of comment about them?

src/transport.ts
```diff
@@ -58,52 +58,39 @@ export class WebRTCTransport implements Transport, Initializable {
...
     // create data channel
+    let dataChannelOpenPromise = defer();
     let handshakeDataChannel = peerConnection.createDataChannel('data', { negotiated: true, id: 1 });
-    //
+    handshakeDataChannel.onopen = (_) => dataChannelOpenPromise.resolve();
+    handshakeDataChannel.onerror = (ev: Event) => {
+      log.error('Error opening a data channel for handshaking: %s', ev.toString());
+      dataChannelOpenPromise.reject(dataChannelError('data', `error opening datachannel: ${ev.toString()}`));
+    };
+    setTimeout(() => {
```
GlenDC, line 75: you never cancel this time out as you do not assign it. So while the promise won't be affected, you will still log the error?!

src/transport.ts
```diff
@@ -58,52 +58,39 @@ export class WebRTCTransport implements Transport, Initializable {
...
+
     // create offer sdp
     let offerSdp = await peerConnection.createOffer();
-    //
-    //
     // generate random string for ufrag
-    let ufrag = genUuid();
-
-    //
+    let ufrag = genUuid().replaceAll('-', '');
```
GlenDC, line 83: is there now ay to ge the uuid without having to replace chars?

src/transport.ts
```diff
@@ -112,7 +99,8 @@ export class WebRTCTransport implements Transport, Initializable {
     //set the Noise Prologue to libp2p-webrtc-noise:<FINGERPRINTS> before starting the actual Noise handshake.
     //  <FINGERPRINTS> is the concatenation of the of the two TLS fingerprints of A and B in their multihash byte representation, sorted in ascending order.
     let fingerprintsPrologue = this.generateNoisePrologue(peerConnection, ma);
-    let noise = new Noise(myPeerId.privateKey, undefined, stablelib, fingerprintsPrologue);
+    let noise = new Noise(undefined, undefined, undefined, fingerprintsPrologue);
```
GlenDC, line 102: Are these undefines really what we want here? Also if that is an expected use case than IMHO that function signature is pretty poor, as you would in JS expect it to be part as an optional dict at the end instead, forcing to put these undefined is a pretty poor choice.

Or is this just a temporary thing on our end? either way, might at the very least want to document why you're putting undefines here

src/transport.ts
```diff
@@ -112,7 +99,8 @@ export class WebRTCTransport implements Transport, Initializable {
...
+    // let noise = new Noise(undefined, undefined, stablelib, fingerprintsPrologue);
```
GlenDC, line 103: Forgot to delete?

src/transport.ts
```diff
@@ -139,16 +127,11 @@ export class WebRTCTransport implements Transport, Initializable {
...
-    if (!remoteCerthash) {
-      throw inappropriateMultiaddr('no remote tls fingerprint in multiaddr');
-    }
-    let remote = base64.decode(remoteCerthash);
     if (pc.getConfiguration().certificates?.length === 0) {
       throw invalidArgument('no local certificate');
     }
-    let localCert = pc.getConfiguration().certificates![0];
-    if (localCert.getFingerprints().length === 0) {
+    let localCert = pc.getConfiguration().certificates?.at(0)!;
```
GlenDC, line 133: OOOoooh I didn't know typescript allows optional chaining :o Need to learn more about it, thanks for showing me!

test/connection.browser.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('connection browser tests', () => {
       break;
     }
     expect(responsed).to.be.true();
-  });
+  }).timeout(54321);
```
GlenDC, line 44: what's the point of such a big time out?

test/transport.browser.spec.ts
```diff
@@ -89,3 +94,29 @@ describe('basic transport tests', () => {
...
+      registrar: mockRegistrar(),
+    });
+    t.init(components);
+    let ma = new Multiaddr(SERVER_MULTIADDR);
+    let conn = await t.dial(ma, ignoredDialOption());
+    let stream = await conn.newStream(['/echo/1.0.0']);
+    let data = 'dataToBeEchoedBackToMe\n';
+    let response = await pipe([uint8arrayFromString(data)], stream, async (source) => await first(source));
+    expect(response?.subarray()).to.equalBytes(uint8arrayFromString(data));
+    // console.log('Response was suppsed to be', data);
```
GlenDC, line 119: forgot to delete?

tsconfig.json
```diff
@@ -1,8 +1,12 @@
 {
 	"extends": "aegir/src/config/tsconfig.aegir.json",
 	"compilerOptions": {
```
GlenDC, line 3: might want to make the ts rules a bit more strict, still allows quiet some potential errors from what I can tell, or perhaps I am wrong. You tell me.

test/server-multiaddr.js
```diff
@@ -0,0 +1 @@
+export var SERVER_MULTIADDR = '';
```
GlenDC, line 1: smells like bad design to me. Can better be handled with an ENV variable, no?

GlenDC (changes requested):
Did not yet go too deep into the business logic itself, as we are also missing some documentation. Next to these comments I suggest to add a README section (for developers) about the testing, how to run them, and what are some special pre-requisites (e.g. the go-libp2p setup and special things there). Also I would make errors more clear where you can. E.g. if you say "cannot run this because I need that, than you might as well explain how to provide 'that'".

Once these comments are resolved and the documentation is also improved, I could try to look into business logic if desired. I am new to libp2p and webrtc though, so that will require some serious learning on the fly :) For that reason might also be useful if you can link in the documentation the relevant specs and RFC's.

test/connection.browser.spec.ts
ddimaria: @John-LittleBearLabs I thought this was removed in your PR?

.gitignore
```diff
@@ -36,6 +36,7 @@ bower_components
 Thumbs.db
 
 # Ignore built ts files
+dist/
```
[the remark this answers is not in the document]
ckousik: It wasn't covering the top level dist folder (although git would ignore it) and was causing all my fuzzy finding tools to pick up js files in dist.

src/connection.ts
ckousik: Yeah, this is an artifact of debugging.

test/server-multiaddr.js
ckousik: @John-LittleBearLabs Should we switch to the http interface on go-libp2p instead of using the file?

src/connection.ts
ckousik: User does not call this method. This only sets up a callback handler for incoming streams.

src/transport.ts
ckousik: good catch! thank you

src/transport.ts
ckousik: We don't really need a uuid, just a random string. Removing the `-` character is a webrtc requirement where the ufrag and password cannot have that character.

src/transport.ts
ckousik: This is the function signature expected by the Noise library: https://github.com/ChainSafe/js-libp2p-noise/blob/master/src/noise.ts#L38
We do not use any of the earlier parameters and it does not accept a dict, so we are forced to put these undefined. We can always make a PR to the Noise library

test/server-multiaddr.js
ckousik: @GlenDC This was a temporary automation for running the interop test. This test will likely be removed in favor of testground.

GlenDC: I understand. If you do not think there is a clearer solution possible without much work, fine for me. But I imagine setting an address can just as well be done via an env variable that you can pass with your cli commands, no? Just an idea though.

src/connection.ts
GlenDC: Even so, can we simply silence it? Does it not need handling in the flow where we might need to do something to act upon besides just logging? Or does the rest of the code flow handle this situation just the same as any?

.gitignore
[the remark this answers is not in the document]
GlenDC: Ah yeah ok, in that case I think you can just remove `dist/**/*` though, fairly certain that is now obsolete given `dist/` should cover the same if I recall

src/connection.ts
ckousik: So, the possible errors in this try block could be as follows:
1. There is no protocol proposed in the incoming stream that we support. 
2. We select an incoming protocol, and there is a failure in the handler which is user-defined.
in either case, we do not want the error to affect the connection object, since the errors may not be network errors.

test/server-multiaddr.js
GlenDC: Fair enough. It's anyway just a test file, so no big deal. Feel free to keep as-is, we already talked way more about it than necessary anyway :D

src/connection.ts
GlenDC: I haven't looked too much into the business logic, but your reasoning seems sound. Thanks for the clarification. I'm still suspicious of the error silencing here, but I'll just keep it in the back of my mind for now. Thank you.

src/sdp.ts
ckousik: Not sure if a comment explaining is useful, but here is the grammar for the candidate line: https://www.rfc-editor.org/rfc/rfc8839.html#name-candidate-attribute . 

The foundation (first number) is a way to encode ice candidates. The RFC specifies what the foundation should do here: https://www.rfc-editor.org/rfc/rfc8445.html#section-5.1.1.3 , but in practice it is not used much. This can be any number for our purposes.

The second number is the stream component. `1` is RTP and `2` is RTCP. Since we multiplex RTP and RTCP over the same component, we use `1`. 

The third number (after UDP) is the candidate priority. Since we have only 1 candidate, this does not matter in this SDP.

GlenDC: What I would do is make constants for this with very explicit declarative names, and you can link to the relevant paragraph in an RFC for each constant. I think that's enough documentation. Gives all the info with little effort.

test/transport.browser.spec.ts
```diff
@@ -89,3 +94,28 @@ describe('basic transport tests', () => {
...
 });
 
+import { SERVER_MULTIADDR } from './server-multiaddr';
+
+describe('Transport interoperability tests', () => {
+  it('can connect to a server', async () => {
+    if (SERVER_MULTIADDR) {
+      console.log('Will test connecting to', SERVER_MULTIADDR);
+    } else {
+      console.log('Will not test connecting to an external server, as we do not appear to have one.');
```
ddimaria, line 104: With `export var SERVER_MULTIADDR = '';`, does this test never run?

ckousik: It gets rewritten through an npm script. cc @John-LittleBearLabs

src/connection.ts
```diff
@@ -60,14 +61,22 @@ export class WebRTCConnection implements ic.Connection {
       },
     };
     this.handleIncomingStreams();
+    this.peerConnection.onconnectionstatechange = (_) => {
```
GlenDC, line 64: why not just put `const closed = ['closed', 'failed', 'disconnected'].includes(this.peer.connectionState);`

Avoids the whole switch statement etc.

ddimaria (approved):
LGTM

GlenDC (approved):

test/transport.browser.spec.ts
John-LittleBearLabs: Yeah. If you're doing test and not test:interop then it doesn't get overwritten and you won't run this.

PR #29, opened 2022-09-28
Upgradable MultiaddrConnection

ckousik:
We use the  new Upgrader interface to use in-built encryption and multiplexing of a WebRTC connection.


src/sdp.ts
```diff
@@ -6,7 +6,7 @@ import { bases } from 'multiformats/basics';
 
 const log = logger('libp2p:webrtc:sdp');
 
-const mbdecoder = (function () {
+export const mbdecoder = (function () {
```
John-LittleBearLabs, line 9: Why's this exported?

src/stream.ts
```diff
@@ -44,7 +45,7 @@ export class WebRTCStream implements Stream {
   metadata: Record<string, any>;
   private readonly channel: RTCDataChannel;
 
-  source: Source<Uint8ArrayList> = pushable();
+  _src: Source<Uint8ArrayList> = pushable();
```
John-LittleBearLabs, line 48: Do I misremember "source" being part of the interface or has the interface changed?

src/stream.ts
```diff
@@ -168,6 +179,9 @@ export class WebRTCStream implements Stream {
     this.readClosed = true;
     this.writeClosed = true;
     this.channel.close();
+    if (this.closeCb) {
```
John-LittleBearLabs, line 182: I had to think for a second about what cb might stand for. Perhaps the name could be more verbose.

src/options.ts
```diff
@@ -6,5 +6,4 @@ export interface WebRTCListenerOptions extends CreateListenerOptions {
   //   channelOptions?: WebRTCReceiverInit
 }
 
-export interface WebRTCDialOptions extends DialOptions {
-}
+export interface WebRTCDialOptions extends DialOptions {}
```
John-LittleBearLabs, line 9: Shouldn't there be something here specifying your upgrader as the default one?

John-LittleBearLabs (approved):

src/stream.ts
ckousik: No, we just use a getter/setter pattern for accessing `source`.

PR #46, opened 2022-10-13
downgrade noise to 8.0.0

ckousik:
Downgrade the noise dependency to 8.0.0 until libp2p makes a new release.


achingbrain:
Libp2p has made a new release!

PR #50, opened 2022-10-17
Changes required(?) for importing

plauche:


plauche:
Sorry about the whitespaces changes. I'll try to clean those up.

achingbrain:
You can also lean on `aegir` for this:

```console
$ npx aegir check-project
```

This will bring the module in to line with the rest of the `@libp2p/*` projects:

1. Make sure `package.json` has the correct config for the project type and is consistently formatted
3. CI files are up to date
4. The README is consistently formatted

The README in this module is a bit weird in that there's a [Contribute](https://github.com/little-bear-labs/js-libp2p-webrtc#contribute) block in the middle of it, just delete that and aegir will add the standard one on the end.

You can also use aegir to apply the same linting standards as every other package:

```console
$ npx aegir lint
```

Though I notice this package has it's own `.eslintrc` so you might want to remove that as the rules clash.

ckousik (changes requested):
Why do we need the internals like SDP to be exported? The webrtc function already exports the transport.

ddimaria:
@plauche did we want this merged before the demo?

plauche:
@ckousik I've trimmed down the changes to the minimum required, including removing those exports which are no longer needed. However I'm not sure how to import the package without building it manually...and I don't think committing the `dist/` folder is the correct way to do it. Maybe this is when we need to publish a release?

ckousik (approved):

PR #51, opened 2022-10-18
Fix noise prologue generation

ckousik:
Fix noise prologue generation to match the webrtc transport spec. Also ensures only sha-256 is used for TLS fingerprints.


John-LittleBearLabs (approved):
lgtm

PR #54, opened 2022-10-31
Length prefixed protobuf messages

ckousik:
Add length prefix to protobuf messages.


ckousik:
not ready to merge

src/transport.ts
```diff
@@ -90,7 +90,7 @@ export class WebRTCTransport implements Transport {
     // create offer sdp
     let offerSdp = await peerConnection.createOffer();
     // generate random string for ufrag
-    const ufrag = genUuid().replaceAll('-', '');
+    const ufrag = "libp2p+webrtc+v1/" + genUuid().replaceAll('-', '');
```
GlenDC, line 93: why is this prefix added?
The comment mentions what the code is doing, but not its intention, which is pretty much the only thing a comment can be useful for (imho)

src/stream.ts
```diff
@@ -231,10 +252,35 @@ export class WebRTCStream implements Stream {
...
       log.error(`Exception while sending flag ${flag}: ${e}`);
     }
   }
 }
+
+class MessageState {
+  public buffer: Uint8Array = new Uint8Array()
+  public messageSize: number = 0;
+
+  public bytesRemaining(): number {
```
GlenDC, line 268: I am not a type script expert, but is this not what you would use getters for?, so that they act like regular (read-only) properties?

src/stream.ts
```diff
@@ -160,11 +155,37 @@ export class WebRTCStream implements Stream {
...
-      this.channel.send(send_buf);
+      const res = buf.subarray();
+      const msgbuf = pb.Message.toBinary({message: buf.subarray()});
+      const sendbuf = lp.encode.single(msgbuf)
+      log.trace(`[stream:${this.id}][${this.stat.direction}] sending message: length: ${res.length} ${res}, encoded through pb as ${msgbuf}`);
+      this.channel.send(sendbuf.subarray())
+    }
+  }
+
+  processIncomingProtobuf(buffer: Uint8Array): { data: Uint8Array | undefined } {
```
GlenDC, line 166: why this weird return structure? e.g. how come not just `Uint8Array | undefined`, what's the point of wrapping it in a struct, if you are always unpacking it anyway?!

src/stream.ts
```diff
@@ -131,6 +102,30 @@ export class WebRTCStream implements Stream {
       let err = (evt as RTCErrorEvent).error;
       this.abort(err);
     };
+
+    const self = this;
+    // reader pipe
+    this.channel.onmessage = async ({data}) => {
+      const res = new Uint8Array(data as ArrayBuffer);
```
GlenDC, line 109: if the length is `0` here, would that not mean that data is some kind of null value as well? Can't we just check that early and return already then, without first alloc'ing an Uint8Array here first?

test/transport.browser.spec.ts
```diff
@@ -118,6 +118,7 @@ describe('Transport interoperability tests', () => {
     let data = 'dataToBeEchoedBackToMe\n';
     let response = await pipe([uint8arrayFromString(data)], stream, async (source) => await first(source));
     expect(response?.subarray()).to.equalBytes(uint8arrayFromString(data));
+    await node.stop();
```
GlenDC, line 121: I suppose this is a bug fix? it is neither mentioned nor documented though?

src/stream.ts
```diff
@@ -160,11 +155,37 @@ export class WebRTCStream implements Stream {
       if (closed || this.writeClosed) {
         break;
       }
-      let res = buf.subarray();
-      let send_buf = pb.Message.toBinary({ message: buf.subarray() });
-      log.trace(`[stream:${this.id}][${this.stat.direction}] sending message: length: ${res.length} ${res}, encoded through pb as ${send_buf}`);
-      this.channel.send(send_buf);
+      const res = buf.subarray();
+      const msgbuf = pb.Message.toBinary({message: buf.subarray()});
+      const sendbuf = lp.encode.single(msgbuf)
```
GlenDC, line 160: remind me of C code bases, it's cute, but if I just see `lp` in a code base I cannot say that it's really intuitive to me that this means length prefix o.o

src/stream.ts
```diff
@@ -148,8 +143,8 @@ export class WebRTCStream implements Stream {
       return;
     }
 
-    let self = this;
-    let closeWriteIterable = {
+    const self = this;
```
GlenDC, line 146: don't we have linters that state something an be const given we do not write to it ever again?

src/stream.ts
```diff
@@ -160,11 +155,37 @@ export class WebRTCStream implements Stream {
...
       }
-      let res = buf.subarray();
-      let send_buf = pb.Message.toBinary({ message: buf.subarray() });
-      log.trace(`[stream:${this.id}][${this.stat.direction}] sending message: length: ${res.length} ${res}, encoded through pb as ${send_buf}`);
-      this.channel.send(send_buf);
+      const res = buf.subarray();
+      const msgbuf = pb.Message.toBinary({message: buf.subarray()});
+      const sendbuf = lp.encode.single(msgbuf)
+      log.trace(`[stream:${this.id}][${this.stat.direction}] sending message: length: ${res.length} ${res}, encoded through pb as ${msgbuf}`);
+      this.channel.send(sendbuf.subarray())
```
GlenDC, line 162: More of a question, but why do you have to call `subarray` all the time? Is it to switch type somehow?

src/stream.ts
```diff
@@ -131,6 +102,30 @@ export class WebRTCStream implements Stream {
       let err = (evt as RTCErrorEvent).error;
       this.abort(err);
     };
+
+    const self = this;
+    // reader pipe
+    this.channel.onmessage = async ({data}) => {
```
GlenDC, line 108: are we planning to expand the message object then with more properties?

src/stream.ts
```diff
@@ -56,6 +59,9 @@ export class WebRTCStream implements Stream {
   closed: boolean = false;
   closeCb?: (stream: WebRTCStream) => void | undefined
 
+  // read state
+  messageState: MessageState = new MessageState();
```
GlenDC, line 63: Why is this created but never used?

src/stream.ts
```diff
@@ -1,13 +1,15 @@
...
+import {pushable} from 'it-pushable';
+import * as lp from 'it-length-prefixed';
+import { pipe } from 'it-pipe';
+import defer, {DeferredPromise} from 'p-defer';
 import merge from 'it-merge';
-import { Uint8ArrayList } from 'uint8arraylist';
-import { fromString } from 'uint8arrays/from-string';
-import { logger } from '@libp2p/logger';
+import {Uint8ArrayList} from 'uint8arraylist';
+import {logger} from '@libp2p/logger';
```
GlenDC, line 10: least important comment of all, but why are we something importing as `{ x }` and sometimes as `{x}`? Super unimportant, but that's what a linter is for, no?

src/stream.ts
```diff
@@ -160,11 +155,37 @@ export class WebRTCStream implements Stream {
...
+  processIncomingProtobuf(buffer: Uint8Array): { data: Uint8Array | undefined } {
+    log.trace(`[stream:${this.id}][${this.stat.direction}] received message: length: ${buffer.length} ${buffer}`);
+    const m = pb.Message.fromBinary(buffer);
+    log.trace(`[stream:${this.id}][${this.stat.direction}] received pb.Message: ${Object.entries(m)}`);
+    switch (m.flag) {
+      case undefined:
+        break; //regular message only
+      case pb.Message_Flag.STOP_SENDING:
+        log.trace('Remote has indicated, with "STOP_SENDING" flag, that it will discard any messages we send.');
+        this.closeWrite();
```
GlenDC, line 175: I trust you it's fine, but given these are controlled separately without changing the object's structure somehow it does appear fragile to me. Or can you re-open again after closing?

src/stream.ts
```diff
@@ -131,6 +102,30 @@ export class WebRTCStream implements Stream {
...
+    // reader pipe
+    this.channel.onmessage = async ({data}) => {
+      const res = new Uint8Array(data as ArrayBuffer);
+      if (res.length == 0) {
+        return
+      }
+      this._innersrc.push(res)
+    };
+
+    this._src = pipe(
```
GlenDC, line 116: this little snippet is cute but it seems needlessly complex. Or am I wrong?
Perhaps I am missing some subtle things not making me appreciate its beauty for what it is.

src/transport.ts
ckousik: This is part of the spec: https://github.com/libp2p/specs/pull/412#discussion_r1007462445. I will add a comment as you suggested.

src/stream.ts
ckousik: The `Uint8ArrayList` can have a single message spread across multiple different `Uint8Array` instances. Calling `subarray` makes them readable as a single `Uint8Array` which is much easier to work with.

src/stream.ts
ckousik: no, that seems unlikely to happen.

src/stream.ts
ckousik: artifact of a previous implementation. Removed now.

src/transport.ts
```diff
@@ -135,6 +139,8 @@ export class WebRTCTransport implements Transport {
     })
 
     const muxerFactory = new DataChannelMuxerFactory(peerConnection)
+    // For outbound connections, the remote is expected to start the noise handshake.
```
John-LittleBearLabs, line 142: :+1: on the comments

src/stream.ts
```diff
@@ -144,40 +147,57 @@ export class WebRTCStream implements Stream {
...
+          case StreamStates.CLOSED:
+            this.close();
+        }
       }
-      let res = buf.subarray();
-      let send_buf = pb.Message.toBinary({ message: buf.subarray() });
-      log.trace(`[stream:${this.id}][${this.stat.direction}] sending message: length: ${res.length} ${res}, encoded through pb as ${send_buf}`);
-      this.channel.send(send_buf);
     }
+    return m;
```
John-LittleBearLabs, line 190: I'm somewhat surprised this isn't `return m.message;`
The caller isn't planning on dealing with the flag again, is it?

John-LittleBearLabs (approved):

mxinden:
@ckousik does this resolve the compatibility issues you were seeing with https://github.com/libp2p/rust-libp2p/pull/2622?

ckousik:
> @ckousik does this resolve the compatibility issues you were seeing with [libp2p/rust-libp2p#2622](https://github.com/libp2p/rust-libp2p/pull/2622)?

yes

PR #60, opened 2022-11-18
Feature/refactor and add tests and coverage

ddimaria:
- [x] Code-level documentation
- [x] Code coverage
- [x] Enhance README
- [x] More tests
- [x] Add browser-to-server example

Also, fixes https://github.com/little-bear-labs/js-libp2p-webrtc/issues/53


src/error.ts
```diff
@@ -1,13 +1,6 @@
 import errCode from 'err-code'
```
ddimaria, line 1: I just alpha sorted the errors

src/error.ts
```diff
@@ -99,25 +87,37 @@ export function operationAborted (context: string, reason: string) {
   return errCode(new OperationAbortedError(context, reason), codes.ERR_ALREADY_ABORTED)
 }
 
-export class DataChannelError extends WebRTCTransportError {
-  constructor (streamLabel: string, errorMessage: string) {
-    super(`[stream: ${streamLabel}] data channel error: ${errorMessage}`)
-    this.name = 'WebRTC/DataChannelError'
+export class OverStreamLimitError extends WebRTCTransportError {
```
ddimaria, line 90: Renamed this for consistency

src/transport.ts
```diff
@@ -19,53 +19,93 @@ import * as sdp from './sdp.js'
...
     const rawConn = await this._connect(ma, options)
     log(`dialing address - ${ma.toString()}`)
     return rawConn
   }
 
+  /**
+   * Create transport listeners
+   */
   createListener (options: CreateListenerOptions): Listener {
     throw unimplemented('WebRTCTransport.createListener')
```
ddimaria, line 73: @John-LittleBearLabs is there a plan to implement this?

src/error.ts
```diff
@@ -99,25 +87,37 @@ export function operationAborted (context: string, reason: string) {
...
+export function overStreamLimit (dir: Direction, proto: string) {
+  const code = dir === 'inbound' ? codes.ERR_TOO_MANY_INBOUND_PROTOCOL_STREAMS : codes.ERR_TOO_MANY_OUTBOUND_PROTOCOL_STREAMS
+  return errCode(new OverStreamLimitError(`${dir} stream limit reached for protocol - ${proto}`), code)
 }
 
-export class StreamingLimitationError extends WebRTCTransportError {
-  constructor (msg: string) {
+export class UnimplementedError extends WebRTCTransportError {
+  constructor (methodName: string) {
+    super('A method (' + methodName + ') was called though it has been intentionally left unimplemented.')
```
GlenDC, line 104: isn't it frowned upon to not use template strings, now that these are widely available?

GlenDC (approved):

src/error.ts
ddimaria: Good call, this pattern is updated throughout this file.

src/transport.ts
ckousik: No, this is not going to be supported for browsers any time soon until the browser-to-browser spec is  ready.

ckousik (commented):
This looks good from a preliminary read. I will review it again shortly.

src/transport.ts
ddimaria: Thanks @ckousik, I made some comments in the README about this when discussing the the implementation of the `Transport` interface:

> Browsers can only `dial`, so `listen` is not supported.

examples/go-libp2p-server/go.mod
```diff
@@ -0,0 +1,111 @@
+module github.com/little-bear-labs/js-libp2p/go-libp2p-server
+
+go 1.18
+
+replace github.com/libp2p/go-libp2p => github.com/ckousik/go-libp2p v0.23.3-0.20221029141116-67b7c6290b11
```
ddimaria, line 5: We will need to update this once the Go PR is merged.  I'll add an issue for it.

examples/browser-to-server/README.md
```diff
@@ -0,0 +1,34 @@
+# js-libp2p-webrtc Browser to Server
+
+This example leverages the [vite bundler](https://vitejs.dev/) to compile and serve the libp2p code in the browser. You can use other bundlers such as Webpack, but we will not be covering them here.
+
+## Running the Go Server
+
+To run the Go LibP2P WebRTC server:
+
+```shell
+npm run go-libp2p-server
```
John-LittleBearLabs, line 10: Am I to understand correctly that this command assumes you have go installed already? And at whatever version is necessary? Is it possible npm could make sure that's true for you?

John-LittleBearLabs (approved):
+1 on the readme stuff.

maybe the go-libp2p-server directory could be nested inside of browser-to-server if that's the only place it's ever used from? Having it directly inside of examples, to me, makes it sound like it's an example of how to use js-libp2p-webrtc.

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
+![https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/interface-transport](https://raw.githubusercontent.com/libp2p/js-libp2p-interfaces/master/packages/interface-transport/img/badge.png)
 
-`libp2p-webrtc` accepts WebRTC encapsulated addresses: `/ip4/0.0.0.0/udp/56093/webrtc/certhash/uEiByaEfNSLBexWBNFZy_QB1vAKEj7JAXDizRs4_SnTflsQ`
+Browsers can only `dial`, so `listen` is not supported.
 
-### Connection
+```js
+interface Transport {
+  [Symbol.toStringTag]: string
+  [symbol]: true
```
GlenDC, line 75: What does this line do?

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
-[![](https://raw.githubusercontent.com/libp2p/js-libp2p-interfaces/master/packages/libp2p-interfaces/src/transport/img/badge.png)](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/libp2p-interfaces/src/transport)
+![https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/interface-transport](https://raw.githubusercontent.com/libp2p/js-libp2p-interfaces/master/packages/interface-transport/img/badge.png)
 
-`libp2p-webrtc` accepts WebRTC encapsulated addresses: `/ip4/0.0.0.0/udp/56093/webrtc/certhash/uEiByaEfNSLBexWBNFZy_QB1vAKEj7JAXDizRs4_SnTflsQ`
+Browsers can only `dial`, so `listen` is not supported.
 
-### Connection
+```js
+interface Transport {
+  [Symbol.toStringTag]: string
```
GlenDC, line 74: After reading up I found out what the implementation of this line would do, but I do not get why it is useful to declare here?

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
+  [symbol]: true
+  dial: (ma: Multiaddr, options: DialOptions) => Promise<Connection>
+  createListener: (options: CreateListenerOptions) => Listener
+  filter: MultiaddrFilter
+}
+
+class WebRTCTransport implements Transport {
+
+  async dial (ma: Multiaddr, options: WebRTCDialOptions): Promise<Connection> {
```
GlenDC, line 83:
```suggestion
  async dial (ma: Multiaddr, options?: WebRTCDialOptions): Promise<Connection> {
```

Given options implies it is optional, not better to make it clear that the object as a whole is optional? Or is there value in enforcing `{}` being passed in when no options are desired?

Or is it because there are properties in there required, but if so, it probably shouldn't be part of options?

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
+    const rawConn = await this._connect(ma, options)
+    log(`dialing address - ${ma.toString()}`)
```
GlenDC, line 85:
```suggestion
    log(`dialing address - ${ma}`)
```

Fairly certain that `toString()` is implicitly called, no?
Tried my theory out and it does seem to be the case:

```
> class Foo { toString() { return "bar" } }
> console.log(`Hello, ${new Foo()}`)
Hello, bar
```

<https://hacks.mozilla.org/2015/05/es6-in-depth-template-strings-2/> seems to agree with this as well.

GlenDC: Also very minor, and perhaps I miss a convention here, but the log says `dialing` but at the point of the log you already `dialed`, no? So either we can simplify it to:

```
log(`dialing address - ${ma}`)
return this._connect(ma, options)
```

or else it perhaps is better to say `dailed`? Or am I missing something here?

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
+    return rawConn
+  }
+
+  createListener (options: CreateListenerOptions): Listener {
```
GlenDC, line 89: if you turn your class into an abstract class (`abstract class WebRTCTransport implements Transport`) you should be able to just do:

```
abstract createListener (options: CreateListenerOptions): Listener
```

without having to specify your dummy body.

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
+    throw unimplemented('WebRTCTransport.createListener')
+  }
+}
+```
 
-[![](https://raw.githubusercontent.com/libp2p/js-libp2p-interfaces/master/packages/libp2p-interfaces/src/connection/img/badge.png)](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/libp2p-interfaces/src/connection)
+### Connection
 
-## Hacking
+![https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/interface-connection](https://raw.githubusercontent.com/libp2p/js-libp2p-interfaces/master/packages/interface-connection/img/badge.png)
```
GlenDC, line 97:
```suggestion
[![Connection interface for libp2p](https://raw.githubusercontent.com/libp2p/js-libp2p-interfaces/master/packages/interface-connection/img/badge.png)](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/interface-connection)
```

Not more accessible?

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
 
-Besides the usual `npm install` to get dependencies, `npm run build` to invoke tsc, and `npm run test` to execute unit tests...
+```js
+interface MultiaddrConnection extends Duplex<Uint8Array> {
+  close: (err?: Error) => Promise<void>
+  remoteAddr: Multiaddr
```
GlenDC, line 102: Does it has to be writable or can it be `readonly`? Same question for the next property.

README.md
```diff
@@ -35,49 +36,81 @@ npm i @libp2p/webrtc
...
+  timeline: MultiaddrConnectionTimeline
+}
 
-There is also `npm run autogen` which uses ProtoBuf's protoc to populate the generated code directory `proto_ts` based on `*.proto` files in src. Don't forget to run this step before `build` any time you make a change to any of the `*.proto` files.
+class WebRTCMultiaddrConnection implements MultiaddrConnection { }
```
GlenDC, line 106: Is there any value to this example as it:

- neither implements the connection
- nor does it get past a decent linter for several reasons

Just asking as people do tend to copy examples as is.

examples/browser-to-server/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+Copy the multiaddress in the output.
```
GlenDC, line 13: might be useful to show an example of such output and multiaddress

examples/browser-to-server/README.md
```diff
@@ -0,0 +1,34 @@
...
+## Running the Example
+
+In a separate console tab, install dependencies and start the Vite server:
+
+```shell
+npm i && npm run start
+```
+
+The browser window will automatically open.
+Using the copied multiaddress from the Go server, pasted that value into the `Server MultiAddress` input and click the `Connect` button.
```
GlenDC, line 24:
```suggestion
Using the copied multiaddress from the Go server, paste it into the `Server MultiAddress` input and click the `Connect` button.
```

examples/browser-to-server/README.md
```diff
@@ -0,0 +1,34 @@
...
+Once the peer is connected, click the message section will appear.  Enter a message and click the `Send` button.
```
GlenDC, line 25:
```suggestion
Once the peer is connected, click the message section once it appeared.  Then enter a message and click the `Send` button.
```
?

examples/browser-to-server/index.js
```diff
@@ -0,0 +1,39 @@
+import { createLibp2p } from 'libp2p'
+import { Noise } from '@chainsafe/libp2p-noise'
+import { multiaddr } from '@multiformats/multiaddr'
+import first from "it-first";
+import { pipe } from "it-pipe";
+import { fromString, toString } from "uint8arrays";
+import { webRTC } from 'js-libp2p-webrtc'
+
+  let stream;
```
GlenDC, line 9: Is this indention on purpose? o.O

examples/browser-to-server/index.js
```diff
@@ -0,0 +1,39 @@
...
+  const sendSection = document.getElementById('send-section')
+  const appendOutput = (line) => output.innerText += `${line}\n`
+  const clean = (line) => line.replaceAll('\n', '')
+
+  const node = await createLibp2p({
+    transports: [webRTC()],
+    connectionEncryption: [() => new Noise()],
+  });
+  
+  await node.start()
```
GlenDC, line 20: Does this even run? Since when can you have `await` in a non-async body, which the global body is AFAIK?

examples/browser-to-server/index.js
GlenDC: Or did you forget to define `(async () => {` on top?

GlenDC: Even better, probably only want to do it once you know the DOM is initialised/ready?

examples/browser-to-server/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "js-libp2p-webrtc-browser-to-server",
+  "version": "1.0.0",
+  "description": "Connect a browser to a server",
+  "type": "module",
+  "scripts": {
+    "start": "vite",
+    "go-libp2p-server": "cd ../go-libp2p-server && go build && ./go-libp2p-server"
```
GlenDC, line 8: does `npm` run your scripts in a new shell? As otherwise this command would change your `CWD` implicitly, which might be confusing and not expected when the cmd exits?

examples/go-libp2p-server/go.mod
GlenDC: Link to the issue in a comment perhaps with a big fat TODO?

examples/go-libp2p-server/main.go
```diff
@@ -0,0 +1,96 @@
...
+func main() {
+	host := createHost()
+	host.SetStreamHandler("/echo/1.0.0", echoHandler)
+	defer host.Close()
+	remoteInfo := peer.AddrInfo{
+		ID:    host.ID(),
+		Addrs: host.Network().ListenAddresses(),
+	}
+
+	remoteAddrs, _ := peer.AddrInfoToP2pAddrs(&remoteInfo)
```
GlenDC, line 73: you know a language has poor error handling when examples usually omit the error handling  :D

examples/go-libp2p-server/main.go
```diff
@@ -0,0 +1,96 @@
...
+	h, err := libp2p.New(
+		libp2p.Transport(webrtc.New),
+		libp2p.ListenAddrStrings(
+			fmt.Sprintf("/ip4/%s/udp/0/webrtc", listenerIp),
+		),
+		libp2p.DisableRelay(),
+		libp2p.Ping(true),
+	)
+	if err != nil {
+		panic(err)
```
GlenDC, line 92: given you already use `log` pkg anyway, you might perhaps want to use `log.Fatal(err)` instead?

src/maconn.ts
```diff
@@ -8,17 +8,46 @@ import { nopSink, nopSource } from './util.js'
...
-  private readonly peerConnection: RTCPeerConnection;
+  /**
+   * WebRTC Peer Connection
+   */
+  readonly peerConnection: RTCPeerConnection;
+
+  /**
+   * The multiaddr address used to communicate with the remote peer
+   */
   remoteAddr: Multiaddr;
```
GlenDC, line 36: does this property and the ones below have to be public? If so, do they have to writable? If so, can they be public only as a `get`(ter)?

src/maconn.ts
```diff
@@ -28,7 +57,11 @@ export class WebRTCMultiaddrConnection implements MultiaddrConnection {
   }
 
   async close (err?: Error | undefined): Promise<void> {
-    log.error('error closing connection', err)
+    if (err !== undefined) {
+      log.error('error closing connection', err)
+    }
```
GlenDC, line 62: might be nice to have a `non-err` `log.debug` equivalent for sanity checking while debugging?

src/muxer.ts
```diff
@@ -7,7 +6,14 @@ import { WebRTCStream } from './stream.js'
 import { nopSink, nopSource } from './util.js'
 
 export class DataChannelMuxerFactory implements StreamMuxerFactory {
+  /**
+   * WebRTC Peer Connection
+   */
   private readonly peerConnection: RTCPeerConnection
+
+  /**
+   * The string representation of the protocol, requried by StreamMuxerFactory
```
GlenDC, line 15:
```suggestion
   * The string representation of the protocol, required by `StreamMuxerFactory`
```

src/muxer.ts
```diff
@@ -19,21 +25,61 @@ export class DataChannelMuxerFactory implements StreamMuxerFactory {
...
+   * WebRTC Peer Connection
+   */
   private readonly peerConnection: RTCPeerConnection
+  /**
+   * The protocol as represented in the multiaddress
+   */
   readonly protocol: string = '/webrtc'
+
+  /**
+   * Array of streams in the data channel
```
GlenDC, line 42: Any streams? Just the ones alive? Or once in the array always in the array?

src/options.ts
```diff
@@ -1,8 +1,4 @@
 import { CreateListenerOptions, DialOptions } from '@libp2p/interface-transport'
 
-export interface WebRTCListenerOptions extends CreateListenerOptions {
-  //, WebRTCInitiatorInit {
-  //   channelOptions?: WebRTCReceiverInit
-}
-
+export interface WebRTCListenerOptions extends CreateListenerOptions {}
```
GlenDC, line 3: my linters do not accept such empty extends, I believe better practice is something like:

```
export { CreateListenerOptions as WebRTCListenerOptions } from '@libp2p/interface-transport'
```

src/sdp.ts
```diff
@@ -76,7 +86,9 @@ export function toSupportedHashFunction (name: multihashes.HashName): string {
   }
 }
 
-// Convert a multiaddr into a SDP
+/**
+ * Convert a multiaddr into a SDP
```
GlenDC, line 90: Might be nice to write fully what `SDP` means as `<full name> (SDP)`, small effort but a cognitive ease of mind

src/sdp.ts
```diff
@@ -99,15 +111,19 @@ a=max-message-size:100000
...
 export function fromMultiAddr (ma: Multiaddr, ufrag: string): RTCSessionDescriptionInit {
   return {
     type: 'answer',
     sdp: ma2sdp(ma, ufrag)
   }
 }
 
-// Replace the ufrag and password values in a SDP
+/**
+ * Replace (munge) the ufrag and password values in a SDP
```
GlenDC, line 125: Might just be me reading functions that I require pre-knowledge for, but I stil feels there is a lot of guess work required on my end to get what this function does without reading the actual implementation.

src/stream.ts
```diff
@@ -12,6 +12,9 @@ import * as pb from '../proto_ts/message.js'
 
 const log = logger('libp2p:webrtc:stream')
 
+/**
+ * Constructs a default StreamStat
```
GlenDC, line 16: might be nice to expand a bit on what these `defaults` entail?

src/transport.ts
```diff
@@ -19,53 +19,93 @@ import * as sdp from './sdp.js'
...
+
+/**
+ * The time to wait, in milliseconds, for the data channel handshake to complete
+ */
 const HANDSHAKE_TIMEOUT_MS = 10000
-const WEBRTC_CODE: number = 280
-const CERTHASH_CODE: number = 466
 
+/**
+ * Created by converting the hexadecimal protocol code to an integer.
```
GlenDC, line 29: but weird they do not just write it as hex literals, but ok, seems you're just following the docs here to the point

src/transport.ts
```diff
@@ -19,53 +19,93 @@ import * as sdp from './sdp.js'
...
+   * Dial a given multiaddr
+   */
   async dial (ma: Multiaddr, options: WebRTCDialOptions): Promise<Connection> {
     const rawConn = await this._connect(ma, options)
     log(`dialing address - ${ma.toString()}`)
     return rawConn
   }
 
+  /**
+   * Create transport listeners no supported by browsers
```
GlenDC, line 70: this sentence reads a big ambiguous and unclear, can you expand it please? E.g. `Creating transports listeners is not supported when used from within a browser`, or something like that

test/maconn.browser.spec.ts
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+
+import { multiaddr } from '@multiformats/multiaddr'
+import { expect } from 'chai'
+import { WebRTCMultiaddrConnection } from './../src/maconn'
+
+describe('Multiaddr Connection', () => {
+  it('can open and close', async () => {
```
GlenDC, line 8: imho it doesn't really test that right? I mean it proves the `close` property changes when you call `close()` but it might as well be done using a dummy property which toggles.

Not better to next to what you already do, also write to it, and have it succeed proving you receive the value, with then the the write or so afterwards doesn't work and you instead get some specific error that indicates the conn is closed and that you also confirm the content is indeed not written.

And same for or instead of the reading part?

test/maconn.browser.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    const peerConnection = new RTCPeerConnection()
+    peerConnection.createDataChannel('whatever', { negotiated: true, id: 91 })
+    const remoteAddr = multiaddr('/ip4/1.2.3.4/udp/1234/webrtc/certhash/uEiAUqV7kzvM1wI5DYDc1RbcekYVmXli_Qprlw3IkiEg6tQ')
+    const maConn = new WebRTCMultiaddrConnection({
+      peerConnection: peerConnection,
+      remoteAddr,
+      timeline: {
```
GlenDC, line 15: why not make this the default value, seems like a sane default if you ask me? Would allow you to omit it here as well

test/stream.browser.spec.ts
```diff
@@ -1,74 +1,68 @@
 import * as underTest from '../src/stream'
 import { expect, assert } from 'chai'
 
+function setup (): { peerConnection: RTCPeerConnection, datachannel: RTCDataChannel, webrtcStream: underTest.WebRTCStream } {
+  const peerConnection = new RTCPeerConnection()
+  const datachannel = peerConnection.createDataChannel('whatever', { negotiated: true, id: 91 })
```
GlenDC, line 6: why not make the data channel id optional and generate a random ID by default? Seems better to me than having to invent strings like this?

GlenDC (changes requested):
Functionality it looks ok, even though that's harder for me to tell given I Am not yet familiar with libp2p and its specs. So my review is more generic, in that context I Requested some changes but feel free to ignore them all, as none of my comments indicate a functional flaw AFAIC

PR #68, opened 2022-12-13
fix: remove uuid dependency

ckousik:
Removes the dependency on uuid. Use a simpler random string generator instead.

### DIFF
--- a/src/transport.ts
+++ b/src/transport.ts
@@ -7,16 +7,15 @@ import * as p from '@libp2p/peer-id'
 import type { Multiaddr } from '@multiformats/multiaddr'
 import * as multihashes from 'multihashes'
 import defer from 'p-defer'
-import { v4 as genUuid } from 'uuid'
 import { fromString as uint8arrayFromString } from 'uint8arrays/from-string'
 import { concat } from 'uint8arrays/concat'
-
 import { dataChannelError, inappropriateMultiaddr, unimplemented, invalidArgument } from './error.js'
 import { WebRTCMultiaddrConnection } from './maconn.js'
 import { DataChannelMuxerFactory } from './muxer.js'
 import type { WebRTCDialOptions } from './options.js'
 import * as sdp from './sdp.js'
 import { WebRTCStream } from './stream.js'
+import {genUfrag} from './util.js'
 
 const log = logger('libp2p:webrtc:transport')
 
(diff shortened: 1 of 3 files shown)


achingbrain:
Linting is failing, you can fix it up with `npm run lint -- --fix`

src/util.ts
```diff
@@ -3,3 +3,6 @@ export const nopSource = {
 }
 
 export const nopSink = async (_: any) => {}
+
+const charset = Array.from('ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/')
```
vmx, line 7: Is there a specific reason to include `+` and `/`? I'd keep it simple and only use alphanumeric characters or perhaps only hex ones. Not having `/` makes it e.g. easier to split off the random bytes of the full ufrag, you can be sure it only contains a single `/` from the `libp2p+webrtc+v1/` prefix.

src/transport.ts
```diff
@@ -142,7 +141,7 @@ export class WebRTCTransport implements Transport {
       dataChannelOpenPromise.reject(dataChannelError('data', error))
     }
 
-    const ufrag = 'libp2p+webrtc+v1/' + genUuid().replaceAll('-', '')
+    const ufrag = 'libp2p+webrtc+v1/' + genUfrag(20)
```
vmx, line 144: Is the new random identifier intentionally shorter (20 bytes) than the old one (32 bytes)?

src/util.ts
ckousik: This is a requirement in the spec: https://github.com/libp2p/specs/blob/master/webrtc/README.md?plain=1#L95

src/transport.ts
ckousik: fixed. thanks for pointing this out.

src/util.ts
vmx: Yes, the prefix is a requirement. But I don't see a requirement that the random characters must contain `+`  and `/`.

ckousik: It's not a hard requirement but the choice of characters is from the default set of characters used to generate a ufrag. I don't see a reason to change that. As far as splitting off the random bytes is concerned, we are guaranteed that there is at least one `/` and that all bytes after it are random.

vmx: Thanks, I wondered where this choice came from, but it's from https://www.rfc-editor.org/rfc/rfc8839#section-5.4, where it states it can be `ice-char` which is exactly that list (https://www.rfc-editor.org/rfc/rfc8839#section-5.1). That makes sense then.

vmx (approved):

achingbrain (approved):

Issue #64, opened 2022-12-06
Export metrics


p-shahi:
### Done Criteria 

Export metrics leveraging the work done in https://github.com/libp2p/js-libp2p/issues/1458

### Motivation

Better observability for performance and benchmarking. See: https://github.com/libp2p/js-libp2p/issues/1458#issuecomment-1323833476 and https://github.com/libp2p/js-libp2p-tcp/pull/223/files

ddimaria:
@p-shahi we can get someone to pick this up.  Alex is in the process of moving this repo to the libp2p org, so the work might be done there.  Someone will confirm when this is started.

ddimaria:
@John-LittleBearLabs is picking this up.

PR #69, opened 2022-12-16
feat: Add metrics for common events

lightsofapollo:
Adds metrics for common error cases (timeoout, dtls handshake failure) and few success cases (stream open / stream close) 

Open questions:

 - [ ] The TCP implementation https://github.com/libp2p/js-libp2p-tcp/blob/b93f4c0a921923401264aa560ae259afc9430a93/src/socket-to-conn.ts#L22 has the ability to prefix individual connection events (such as for close or perhaps the muxer?). I was not certain this would be helpful in this context and omitted it here. Should we add it?
 - [ ] Testing metrics collection I'd love to see metrics collection be part of the test suite. I was thinking perhaps either a mock object or an interface which allows collecting metrics in memory (similar to the prom client) but tests can verify the calls later. 

Closes: https://github.com/libp2p/js-libp2p-webrtc/issues/64

### DIFF
--- a/src/maconn.ts
+++ b/src/maconn.ts
@@ -1,4 +1,5 @@
 import type { MultiaddrConnection, MultiaddrConnectionTimeline } from '@libp2p/interface-connection'
+import type { CounterGroup } from '@libp2p/interface-metrics'
 import { logger } from '@libp2p/logger'
 import type { Multiaddr } from '@multiformats/multiaddr'
 import type { Source, Sink } from 'it-stream-types'
@@ -22,6 +23,11 @@ interface WebRTCMultiaddrConnectionInit {
    * Holds the relevant events timestamps of the connection
    */
   timeline: MultiaddrConnectionTimeline
+
+  /**
+   * Optional metrics counter group for this connection
+   */
+  metrics: CounterGroup | null
 }
 
 export class WebRTCMultiaddrConnection implements MultiaddrConnection {
@@ -39,6 +45,11 @@ export class WebRTCMultiaddrConnection implements MultiaddrConnection {
    * Holds the lifecycle times of the connection
    */
   timeline: MultiaddrConnectionTimeline;
+
+  /**
+   * Optional metrics counter group for this connection
+   */
+  metrics?: CounterGroup
 
   /**
    * The stream source, a no-op as the transport natively supports multiplexing
@@ -62,6 +73,7 @@ export class WebRTCMultiaddrConnection implements MultiaddrConnection {
     }
 
     this.timeline.close = new Date().getTime()
+    this.metrics?.increment({ close: true })
     this.peerConnection.close()
   }
 }
--- a/src/transport.ts
+++ b/src/transport.ts
@@ -17,6 +17,7 @@ import { DataChannelMuxerFactory } from './muxer.js'
 import type { WebRTCDialOptions } from './options.js'
 import * as sdp from './sdp.js'
 import { WebRTCStream } from './stream.js'
+import type { CounterGroup, Metrics } from '@libp2p/interface-metrics'
 
 const log = logger('libp2p:webrtc:transport')
 
@@ -42,9 +43,13 @@ export const CERTHASH_CODE: number = 466
 /**
  * The peer for this transport
  */
-// @TODO(ddimaria): seems like an unnessary abstraction, consider removing
 export interface WebRTCTransportComponents {
   peerId: PeerId
+  metrics?: Metrics
+}
+
+export interface WebRTCMetrics {
+  dialerEvents: CounterGroup
 }
 
 export class WebRTCTransport implements Transport {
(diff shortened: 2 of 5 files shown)


lightsofapollo:
NOTE: I've taken the exact pattern mentioned in the TCP listener. We may want to expand certain metrics in future for WebRTC as there are some complex error conditions not captured in this initial draft.

src/transport.ts
```diff
@@ -45,16 +46,31 @@ export const CERTHASH_CODE: number = 466
 // @TODO(ddimaria): seems like an unnessary abstraction, consider removing
 export interface WebRTCTransportComponents {
   peerId: PeerId
+  metrics?: Metrics
```
ddimaria, line 49: @lightsofapollo now that there is more than one attribute to this interface, can you remove my `todo`?

src/transport.ts
```diff
@@ -45,16 +46,31 @@ export const CERTHASH_CODE: number = 466
...
    */
   private readonly components: WebRTCTransportComponents
+  private readonly metrics?: WebRTCMetrics
 
   constructor (components: WebRTCTransportComponents) {
     this.components = components
+
+    if (components.metrics != null) {
+      this.metrics = {
+        dialerEvents: components.metrics.registerCounterGroup('libp2p_webrtc_dialer_errors_total', {
```
ddimaria, line 68: On the TCP version, `errors` is replaced with `events` in the 2 lines.  Are these errors?

ddimaria:
In the TCP version, [metrics are added to the multiaddress connection](https://github.com/libp2p/js-libp2p-tcp/blob/b93f4c0a921923401264aa560ae259afc9430a93/src/index.ts#L105-L110).  Do we want that [here](https://github.com/libp2p/js-libp2p-webrtc/blob/main/src/transport.ts#L188-L194) as well?

src/transport.ts
lightsofapollo: resolved

src/transport.ts
```diff
@@ -42,19 +43,33 @@ export const CERTHASH_CODE: number = 466
...
+  private readonly metrics?: WebRTCMetrics
 
   constructor (components: WebRTCTransportComponents) {
     this.components = components
+
+    if (components.metrics != null) {
+      this.metrics = {
+        dialerEvents: components.metrics.registerCounterGroup('libp2p_webrtc_dialer_events_total', {
+          label: 'event',
+          help: 'Total count of WebRTC dial errors by type'
```
ddimaria, line 69: Should this `errors` be `events`?

src/transport.ts
lightsofapollo: name has been changed

lightsofapollo:
@ddimaria good question about the multiaddr ... I read through that file in the webrtc implementation and it wasn't entirely clear if we need metrics there except in the case of close https://github.com/libp2p/js-libp2p-webrtc/blob/main/src/maconn.ts (see close) wdyt?

ddimaria:
We may want to include the `close` event in `maconn.ts`.  Thoughts?

ddimaria:
There are events related to the `datachannel`.  We may want to include those in `muxer.ts`.

lightsofapollo:
@ddimaria I like that suggestion... The muxer is more related to the number of inbound/outbound connections which I like. I'll add that next.

src/transport.ts
lightsofapollo: fixed

test/maconn.browser.spec.ts
```diff
@@ -11,6 +11,7 @@ describe('Multiaddr Connection', () => {
     const remoteAddr = multiaddr('/ip4/1.2.3.4/udp/1234/webrtc/certhash/uEiAUqV7kzvM1wI5DYDc1RbcekYVmXli_Qprlw3IkiEg6tQ')
     const maConn = new WebRTCMultiaddrConnection({
       peerConnection: peerConnection,
+      metrics: null,
```
lightsofapollo, line 14: Would be nice if we explicitly tested metrics inteface?

src/maconn.ts
```diff
@@ -22,6 +23,11 @@ interface WebRTCMultiaddrConnectionInit {
    * Holds the relevant events timestamps of the connection
    */
   timeline: MultiaddrConnectionTimeline
+
+  /**
+   * Optional metrics counter group for this connection
+   */
+  metrics: CounterGroup | null
```
achingbrain, line 30: I don't think we should accept `null` as a value, better to mark this optional

```suggestion
  metrics?: CounterGroup
```

test/maconn.browser.spec.ts
achingbrain: The [sinon-ts](https://www.npmjs.com/package/sinon-ts) module can be used to do this in a straightforward way.

lightsofapollo:
@achingbrain ❤️ Thanks! I'll have the next iteration up shortly which will be ready to review :)

PR #82, opened 2023-01-14
chore: replace err-code with CodeError

tabcat:
Related: https://github.com/libp2p/js-libp2p/issues/1269

- deps: bump @libp2p/interfaces
- deps: remove err-code
- [x] chore: replace err-code with CodeError


tabcat:
@wemeetagain `src/error.ts` contains all usage of errcode. the usage there makes replacement more difficult. looking for some insight before pushing changes.

All of the errors and error classes are exported but having issues finding where they are used. Doesn't seem like they are used in the source and they don't seem to be in the package exports.

My initial thought is to just rewrite the classes to functions returning strings and combine them when necesary. It seems like the only things that must be the same in the end is the CodeError's message and name (the code must be the same too but this out of the scope of this change).

tabcat:
@wemeetagain I think this solution is pretty good. There is a breaking change but might not be relevant.

src/error.ts
```diff
@@ -14,110 +14,109 @@ export enum codes {
...
+    super(`${dir} stream limit reached for protocol - ${proto}`, code)
     this.name = 'WebRTC/OverStreamLimitError'
   }
 }
 
 export function overStreamLimit (dir: Direction, proto: string) {
-  const code = dir === 'inbound' ? codes.ERR_TOO_MANY_INBOUND_PROTOCOL_STREAMS : codes.ERR_TOO_MANY_OUTBOUND_PROTOCOL_STREAMS
-  return errCode(new OverStreamLimitError(`${dir} stream limit reached for protocol - ${proto}`), code)
+  return new OverStreamLimitError(dir, proto)
 }
```
tabcat, line 100: Breaking change has to do with exported `OverStreamLimitError` parameter change.

tabcat:
Wondering if the Error returning Functions are necessary. They could just be removed as they don't do much now.

src/error.ts
```diff
@@ -14,110 +14,109 @@ export enum codes {
...
 
-export class WebRTCTransportError extends Error {
-  constructor (msg: string) {
-    super(`WebRTC transport error: ${msg}`)
+export class WebRTCTransportError extends CodeError {
+  constructor (msg: string, code: string) {
+    super(`WebRTC transport error: ${msg}`, code)
     this.name = 'WebRTCTransportError'
   }
 }
```
tabcat, line 22: Other breaking change

tabcat:
Lint is failing from a bug in esquery 1.4.1 https://github.com/estools/esquery/issues/135

p-shahi:
@tabcat I merged the aegir PR which I think has the fix for the esquery issue.

ddimaria:
@tabcat can you resolve the conflicts?  I think Chinmay needed to update the return type related to the aegir update, so it's probably a comb on "yours" and "theirs".

src/error.ts
ddimaria: Can we avoid the breaking change by making the parameter optional?

src/error.ts
ddimaria: Same comment here, we could have a default value for `proto` param in the `OverStreamLimitError` constructor (e.g. `unknown`).  We could add TODOs and make the breaking changes with a planned major version increase.

tabcat: I can make proto in `overStreamLimit` optional; note this would be adding a change.
The parameter change I made has to do with the `OverStreamLimitError` which completely changes the parameters.
I'll undo the breaking parts and add a note.

src/error.ts
tabcat: empty string fine?

p-shahi:
@ddimaria @ckousik looks like @tabcat 's resolved conflicts. Let's try to get this merged by end of week?

ddimaria:
> @ddimaria @ckousik looks like @tabcat 's resolved conflicts. Let's try to get this merged by end of week?

There are [lint errors](https://github.com/libp2p/js-libp2p-webrtc/actions/runs/4246118384/jobs/7484394849) in CI.  @tabcat can you resolve them?  If not, I can tackle this today @p-shahi

tabcat:
My mistake, I'll take care of this right now

PR #84, opened 2023-01-20
Review comments

MarcoPolo:
Overall this looks good. I added some changes and left a couple of questions in the code. Once those questions are addressed and my changes make sense, I can update the code and merge these changes.

I ran eslint on the files I touched. It may be better to review with the `?w=1` option to ignore whitespace changes.


src/stream.ts
```diff
@@ -177,246 +188,265 @@ export class WebRTCStream implements Stream {
...
+        return
+      }
+
+      const msgbuf = pb.Message.toBinary({ message: buf.subarray() })
+      const sendbuf = lengthPrefixed.encode.single(msgbuf)
+
+      this.channel.send(sendbuf.subarray())
+    }
+
+    // Close the channel after the source has ended
```
MarcoPolo, line 323: This tripped me up as well. See https://github.com/libp2p/js-libp2p-webtransport/pull/23

src/muxer.ts
```diff
@@ -101,8 +101,10 @@ export class DataChannelMuxer implements StreamMuxer {
   /**
    * Initiate a new stream with the given name. If no name is
    * provided, the id of the stream will be used.
+   * TODO does this comment make sense? The spec says the label SHOULD be an empty string
    */
   newStream (name: string = ''): Stream {
+   // TODO The spec says the label SHOULD be an empty string: https://github.com/libp2p/specs/blob/master/webrtc/README.md#rtcdatachannel-label
```
ddimaria, line 107: I see your point, though the goal was to keep the implementation flexible for future usage:
> This allows future versions of this specification to make use of the RTCDataChannel label property.
I don't have a strong opinion here.  I don't believe this function is invoked directly by implementors.  As is, this seems to comply with the spec, but we could remove the parameter to be more formal about the intention?

src/stream.ts
```diff
@@ -109,8 +119,8 @@ class StreamState {
         case pb.Message_Flag.RESET:
           this.state = StreamStates.CLOSED
           break
-
-        // no default
+        default:
+          unreachableBranch(flag)
```
ddimaria, line 123: 💯

src/muxer.ts
MarcoPolo: > As is, this seems to comply with the spec, but we could remove the parameter to be more formal about the intention?

Agree. We should remove this param.

src/stream.ts
```diff
@@ -177,246 +188,265 @@ export class WebRTCStream implements Stream {
...
+   */
+  closeRead (): void {
+    const [currentState, nextState] = this.streamState.transition({ direction: 'outbound', flag: pb.Message_Flag.STOP_SENDING })
+    if (currentState === nextState) {
+      // No change, no op
+      return
+    }
+
+    if (currentState === StreamStates.OPEN || currentState === StreamStates.WRITE_CLOSED) {
+      this._sendFlag(pb.Message_Flag.STOP_SENDING)
```
ddimaria, line 386: Interesting, I wonder how that semicolon got past the linter 🤔

ddimaria:
Thanks for the PR @MarcoPolo!  The refactors make the code more readable 💯.  @ckousik will respond to 2 of the TODOs on Monday.

src/muxer.ts
ckousik: The param is optional and is from https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-stream-muxer/src/index.ts#L30-L34 . Removing it should be okay.

ckousik:
@MarcoPolo I've addressed the changes in https://github.com/libp2p/js-libp2p-webrtc/tree/ckousik/review . Can I push to the branch on this PR instead?

src/stream.ts
ckousik: The referenced PR and the current implementation would both break on cases similar to:
```
const response = await pipe([request], stream, sink)
// process response
await pipe([request1], stream)
```
However, it might be better to use it-pushable or it-handshake in these cases.
I've refactored the sink to be more readable, and also limited `sink` to being called once similar to the referenced PR.

MarcoPolo: > The referenced PR and the current implementation would both break on cases similar to...

Yes. That's intentional in the js-libp2p design. And the workaround is using a pushable like you mentioned. I'm not a fan of this, but this is how everything else is setup, so it would be even more confusing if this deviated from it.

MarcoPolo:
@ckousik please do

src/stream.ts
```diff
@@ -177,246 +188,273 @@ export class WebRTCStream implements Stream {
...
+  }
+
+  /**
+   * Close the stream for writing, and indicate to the remote side this is being done 'abruptly'
+   *
+   * @see this.closeWrite
+   */
+  reset (): void {
+    // TODO Why are you resetting the stat here?
+    this.stat = defaultStat(this.stat.direction)
```
MarcoPolo, line 428: I think we should remove this. Am I missing something?

MarcoPolo:
Merging this as my repo review comments. @achingbrain this repo is ready for your review whenever you're ready.

Issue #86, opened 2023-01-24
Test example in CI


MarcoPolo:
We should test that the example works in all browsers as part of the CI job.

Here's js-libp2p-webtransport's workflow for the example: https://github.com/libp2p/js-libp2p-webtransport/blob/main/.github/workflows/examples.yml

and the test: https://github.com/libp2p/js-libp2p-webtransport/blob/main/examples/fetch-file-from-kubo/tests/test.js

p-shahi:
@ckousik would you like to take this on?

ckousik:
@p-shahi I'll be prioritizing circuit v2 right now. @ddimaria could help with prioritizing different tasks.

ddimaria:
@MarcoPolo we may be limited with Go testing until the go-libp2p-webrtc PR is merged.  Do you think replacing that target with the Rust implementation would work (and is ready)?  We had an integration test in the repo previously, but were told to leverage Testground instead (and its successor).

MarcoPolo:
>  We had an integration test in the repo previously

This isn't an integration test in the sense that we want to make sure `js-libp2p-webrtc` works. This should be to make sure we don't break the example, and that folks spinning up the example will have it work for them.

>  Do you think replacing that target with the Rust implementation would work

Maybe. The README.md mentions the Go server, so I would just stick to that.

I don't mind picking this up. I don't think it'll take very long. I made this issue simply to track, not to suggest LBL take it :)

ddimaria:
@MarcoPolo I'm working on this now.

p-shahi:
done in https://github.com/libp2p/js-libp2p-webrtc/pull/87

eng-cc:
> Rust implementation

If we have Rust implementation, can we also add it to the example?

ddimaria:
This issue is related to adding the existing example to CI (you may want to move this to a new issue if you want it tracked).  We haven't planned on adding a Rust implementation to the examples as the goal wasn't full interop but rather how to use the JS API.  That said, if you would like to create an issue for adding Rust and submit a PR, we'd happily merge it.

PR #87, opened 2023-01-30
Add playwright tests and add to CI

ddimaria:
Addresses issue https://github.com/libp2p/js-libp2p-webrtc/issues/86

- [x] Add Playwright test for the browser-to-server example
- [x] Integrate Playwright tests in CI

### DIFF
--- a/examples/browser-to-server/index.js
+++ b/examples/browser-to-server/index.js
@@ -3,7 +3,7 @@ import { noise } from '@chainsafe/libp2p-noise'
 import { multiaddr } from '@multiformats/multiaddr'
 import { pipe } from "it-pipe";
 import { fromString, toString } from "uint8arrays";
-import { webRTC } from 'js-libp2p-webrtc'
+import { webRTC } from '@libp2p/webrtc'
 import { pushable } from 'it-pushable';
 
 let stream;
--- a/examples/browser-to-server/index.js
+++ b/examples/browser-to-server/index.js
@@ -9,7 +9,11 @@ import { pushable } from 'it-pushable';
 let stream;
 const output = document.getElementById('output')
 const sendSection = document.getElementById('send-section')
-const appendOutput = (line) => output.innerHTML += `<div>${line}</div>`
+const appendOutput = (line) => {
+  const div = document.createElement("div")
+  div.appendChild(document.createTextNode(line))
+  output.append(div)
+}
 const clean = (line) => line.replaceAll('\n', '')
 const sender = pushable()
 
--- a/examples/browser-to-server/index.js
+++ b/examples/browser-to-server/index.js
@@ -9,7 +9,7 @@ import { pushable } from 'it-pushable';
 let stream;
 const output = document.getElementById('output')
 const sendSection = document.getElementById('send-section')
-const appendOutput = (line) => output.innerText += `${line}\n`
+const appendOutput = (line) => output.innerHTML += `<div>${line}</div>`
 const clean = (line) => line.replaceAll('\n', '')
 const sender = pushable()
 
@@ -27,7 +27,7 @@ node.connectionManager.addEventListener('peer:connect', (connection) => {
 
 window.connect.onclick = async () => {
   const ma = multiaddr(window.peer.value)
-  appendOutput(`Dialing ${ma}`)
+  appendOutput(`Dialing '${ma}'`)
   stream = await node.dialProtocol(ma, ['/echo/1.0.0'])
   pipe(sender, stream, async (src) => {
     for await(const buf of src) {
--- a/examples/browser-to-server/package.json
+++ b/examples/browser-to-server/package.json
@@ -6,14 +6,14 @@
   "scripts": {
     "start": "vite",
     "build": "vite build",
-    "go-libp2p-server": "cd ../go-libp2p-server && go build && ./go-libp2p-server",
+    "go-libp2p-server": "cd ../go-libp2p-server && go run ./main.go",
     "test": "npm run build && playwright test tests"
   },
   "dependencies": {
     "@chainsafe/libp2p-noise": "^11.0.0",
     "@multiformats/multiaddr": "^11.0.5",
     "it-pushable": "^3.1.0",
-    "js-libp2p-webrtc": "../../",
+    "js-libp2p-webrtc": "@libp2p/webrtc",
     "libp2p": "^0.42.0",
     "vite": "^3.1.0"
   },
--- a/examples/browser-to-server/package.json
+++ b/examples/browser-to-server/package.json
@@ -13,7 +13,7 @@
     "@chainsafe/libp2p-noise": "^11.0.0",
     "@multiformats/multiaddr": "^11.0.5",
     "it-pushable": "^3.1.0",
-    "js-libp2p-webrtc": "@libp2p/webrtc",
+    "@libp2p/webrtc": "^1.0.3",
     "libp2p": "^0.42.0",
     "vite": "^3.1.0"
   },
--- a/examples/browser-to-server/package.json
+++ b/examples/browser-to-server/package.json
@@ -5,14 +5,20 @@
   "type": "module",
   "scripts": {
     "start": "vite",
-    "go-libp2p-server": "cd ../go-libp2p-server && go build && ./go-libp2p-server"
+    "build": "vite build",
+    "go-libp2p-server": "cd ../go-libp2p-server && go build && ./go-libp2p-server",
+    "test": "npm run build && playwright test tests"
   },
   "dependencies": {
     "@chainsafe/libp2p-noise": "^11.0.0",
     "@multiformats/multiaddr": "^11.0.5",
-    "js-libp2p-webrtc": "../../",
     "it-pushable": "^3.1.0",
+    "js-libp2p-webrtc": "../../",
     "libp2p": "^0.42.0",
     "vite": "^3.1.0"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.30.0",
+    "test-util-ipfs-example": "^1.0.2"
   }
 }
--- a/examples/browser-to-server/tests/test.js
+++ b/examples/browser-to-server/tests/test.js
@@ -1,0 +1,98 @@
+/* eslint-disable no-console */
+import { test, expect } from '@playwright/test'
+import { playwright } from 'test-util-ipfs-example'
+import { spawn, exec } from 'child_process'
+import { existsSync } from 'fs'
+
+// Setup
+const play = test.extend({
+  ...playwright.servers()
+})
+
+async function spinUpGoLibp2p() {
+  if (!existsSync('../../examples/go-libp2p-server/go-libp2p-server')) {
+    await new Promise((resolve, reject) => {
+      exec('go build',
+        { cwd: '../../examples/go-libp2p-server' },
+        (error, stdout, stderr) => {
+          if (error) {
+            reject(error)
+            console.error(`exec error: ${error}`)
+            return
+          }
+          resolve()
+        })
+    })
+  }
+
+  const server = spawn('./go-libp2p-server', [], { cwd: '../../examples/go-libp2p-server', killSignal: 'SIGINT' })
+  server.stderr.on('data', (data) => {
+    console.log(`stderr: ${data}`, typeof data)
+  })
+  const serverAddr = await (new Promise(resolve => {
+    server.stdout.on('data', (data) => {
+      console.log(`stdout: ${data}`, typeof data)
+      if (data.includes('p2p addr:  ')) {
+        // Parse the addr out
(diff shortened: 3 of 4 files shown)


examples/browser-to-server/index.js
```diff
@@ -9,7 +9,7 @@ import { pushable } from 'it-pushable';
 let stream;
 const output = document.getElementById('output')
 const sendSection = document.getElementById('send-section')
-const appendOutput = (line) => output.innerText += `${line}\n`
+const appendOutput = (line) => output.innerHTML += `<div>${line}</div>`
```
MarcoPolo, line 12: Why this change? Isn't this unnecessarily a XSS attack vector now? It's an example, yes. But people will copy this.

MarcoPolo (changes requested):
One small thing. Otherwise looks good

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,98 @@
...
+import { playwright } from 'test-util-ipfs-example'
+import { spawn, exec } from 'child_process'
+import { existsSync } from 'fs'
+
+// Setup
+const play = test.extend({
+  ...playwright.servers()
+})
+
+async function spinUpGoLibp2p() {
```
ckousik, line 12: nit: spawnGoLibp2p?

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,98 @@
...
+  if (!existsSync('../../examples/go-libp2p-server/go-libp2p-server')) {
+    await new Promise((resolve, reject) => {
+      exec('go build',
+        { cwd: '../../examples/go-libp2p-server' },
+        (error, stdout, stderr) => {
+          if (error) {
+            reject(error)
+            console.error(`exec error: ${error}`)
+            return
```
ckousik, line 21: should this throw an error?

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,98 @@
...
+  const server = spawn('./go-libp2p-server', [], { cwd: '../../examples/go-libp2p-server', killSignal: 'SIGINT' })
+  server.stderr.on('data', (data) => {
+    console.log(`stderr: ${data}`, typeof data)
+  })
+  const serverAddr = await (new Promise(resolve => {
+    server.stdout.on('data', (data) => {
+      console.log(`stdout: ${data}`, typeof data)
+      if (data.includes('p2p addr:  ')) {
+        // Parse the addr out
+        resolve((String(data)).match(/p2p addr:  ([^\s]*)/)[1])
```
ckousik, line 37:
```
        const addr = String(data).match(/p2p addr: ([^\s]*)/)
        if (addr.length === 0) {
          <throw error>
        }
        resolve(addr[1])
```

examples/browser-to-server/index.js
ddimaria: Good call, this has refactored to be less permissive.

examples/browser-to-server/tests/test.js
ddimaria: Good name change, this has been updated.

examples/browser-to-server/tests/test.js
ddimaria: I just checked the various approached and throwing the error results in better error output.  Thanks @ckousik!

examples/browser-to-server/tests/test.js
ddimaria: I had to change the suggestion as we get (and expect) output that doesn't include the multiaddr that isn't an error case.  Additionally, we need to check for the null case when there are no matches.

```
const addr = String(data).match(/p2p addr:  ([^\s]*)/)
if (addr !== null && addr.length > 0) {
  resolve(addr[1])
}   
```

examples/browser-to-server/package.json
```diff
@@ -5,14 +5,20 @@
   "type": "module",
   "scripts": {
     "start": "vite",
-    "go-libp2p-server": "cd ../go-libp2p-server && go build && ./go-libp2p-server"
+    "build": "vite build",
+    "go-libp2p-server": "cd ../go-libp2p-server && go build && ./go-libp2p-server",
```
GlenDC, line 9: Why actually going through the trouble of building it and running it, instead of just using `go run ./main.go` or w/e the binary entry is?

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,96 @@
...
+
+// Setup
+const play = test.extend({
+  ...playwright.servers()
+})
+
+async function spawnGoLibp2p() {
+  if (!existsSync('../../examples/go-libp2p-server/go-libp2p-server')) {
+    await new Promise((resolve, reject) => {
+      exec('go build',
```
GlenDC, line 15: same here, might as well just always run it, no? instead of building it first?

examples/browser-to-server/package.json
```diff
@@ -5,14 +5,20 @@
...
+    "test": "npm run build && playwright test tests"
   },
   "dependencies": {
     "@chainsafe/libp2p-noise": "^11.0.0",
     "@multiformats/multiaddr": "^11.0.5",
-    "js-libp2p-webrtc": "../../",
     "it-pushable": "^3.1.0",
+    "js-libp2p-webrtc": "../../",
```
GlenDC, line 16: actual package is now called @libp2p/webrtc, so we might as well call it here this as well.

examples/browser-to-server/package.json
```diff
@@ -5,14 +5,20 @@
   "type": "module",
```
GlenDC, line 5: I think it falls out of scope for this PR, but something to think about is that we might as well turn this repo into a workspace. This way the example and webrtc package could be both living in the same package, allowing the example to make use of the @libp2p/webrtc package the same as any external usage would. Making the example also directly copyable without having to rely on relative paths in its package.json to make it work given that the relative paths would be fixed by the root workspace definition :) Just my 2 cents on this matters, but feels cleaner to me.

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,96 @@
...
+        resolve(addr[1])
+      }        
+    })
+  }))
+  return { server, serverAddr }
+}
+
+play.describe('bundle ipfs with parceljs:', () => {
+  // DOM
+  const connectBtn = '#connect'
```
GlenDC, line 44: 2 comments:

- is there no convention for constants, in many language it's something like `UPPER_CASE` but dunno about ts or PL in general. Either way I would expect it to be different from regular variable names?
- might want to use a convention suffix for these ones to make clear that these are Html Element Identifier strings rather than actual "buttons", "addresses", "input fields", etc...?

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,96 @@
...
+  const output = '#output'
+
+  let server
+  let serverAddr
+
+  // eslint-disable-next-line no-empty-pattern
+  play.beforeAll(async ({ }, testInfo) => {
+    testInfo.setTimeout(5 * 60_000)
+    const s = await spawnGoLibp2p()
+    server = s.server
```
GlenDC, line 57: dunno what the convention is but unpacking `{ server, serverAddress} = s` should work here, no?

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,96 @@
...
+    serverAddr = s.serverAddr
+    console.log('Server addr:', serverAddr)
+  }, {})
+
+  play.afterAll(() => {
+    server.kill('SIGINT')
+  })
+
+  play.beforeEach(async ({ servers, page }) => {
+    console.log(`http://localhost:${servers[0].port}/`)
```
GlenDC, line 67: When I log and use such a composed value I usually assign it to a const first, but no biggie

examples/browser-to-server/tests/test.js
```diff
@@ -0,0 +1,96 @@
...
+    
+    // add the go libp2p multiaddress to the input field and submit
+    await page.fill(connectAddr, serverAddr)
+    await page.click(connectBtn)
+
+    // send the relay message to the go libp2p server
+    await page.fill(messageInput, message)
+    await page.click(sendBtn)
+
+    await page.waitForSelector('#output:has(div)')
```
GlenDC, line 82: again no biggie but I notice a mix of single and double quotes in this file for no good reason, but given the linter doesn't complain I suppose there is no strict def to be used here?

examples/browser-to-server/tests/test.js
MarcoPolo: >  convention suffix for these ones to make clear that these are Html Element Identifier strings rather than actual "buttons"

That's what the types are for :)

examples/browser-to-server/tests/test.js
MarcoPolo: Good question! this should be documented. `go run ./main.go` builds the program then starts a subprocess for the program. SIGINT on the parent process doesn't stop the subprocess (try running `go run` in the background and issuing a `kill -2 pid`). By running the binary directly we avoid the indirection and SIGINT works as expected.

I wonder if we could SIGKILL this instead and it would work?

I don't think it's worth changing in this PR.

examples/browser-to-server/tests/test.js
MarcoPolo: No. This is assigning to closured variables.

examples/browser-to-server/package.json
MarcoPolo, line 16:
```suggestion
    "@libp2p/webrtc": "../../",
```

examples/browser-to-server/tests/test.js
GlenDC: I agree that it is probably out of scope. More of a note.

examples/browser-to-server/tests/test.js
GlenDC: What types? O.O

GlenDC: You mean that this could be implied because it is of type str? Dunno, not entirely sure that is the case. Certainly not for the output one imho. I like to be very clear about this thing, but neither do I want to make a foothold here, so I'm fine with whatever you end up choosing here, just  note from my POV.

examples/browser-to-server/tests/test.js
GlenDC: Hmmmm, ok.

examples/browser-to-server/tests/test.js
MarcoPolo: Oh my mistake, I thought this was a `.ts` file

examples/browser-to-server/package.json
ddimaria: Thanks Glen, I added this change.

examples/browser-to-server/package.json
ddimaria: The pathing has been removed in favor of the npm package import.  The example was written before the package was released, so this was something that needed to be done now that it's published.  Thanks for the reminder!  

Regarding the placement of this comment, is there a concern with `"type": "module"`?

examples/browser-to-server/tests/test.js
ddimaria: This was a borrow from the WebTransport Playwright test, though Hungarian(ish) notation may be beneficial here.  I'm OK leaving as is.

examples/browser-to-server/package.json
GlenDC: No no, just decided to put it on top of the file it was related to. Sorry for the confusion there.

examples/browser-to-server/tests/test.js
ddimaria: Yes, these are singletons where the variables are scoped in the `play.describe` closure.  The styleguide for this repo doesn't use semicolons, so doing the recommended destructuging assignment on an already defined variable causes issues:
```
({ server, serverAddr } = await spawnGoLibp2p())
```
The parser appends this to the `setTimeout()` function call above it (even with an extra line break), resulting in a TypeError: `TypeError: testInfo.setTimeout(...) is not a function`.

Adding semicolons resolves the issue but makes the code incompatible with the main `js-libp2p` styles and one that Aegir expects in lints.

I'm going to leave as is.

GlenDC: More then fair enough. Sorry for the trouble 😰

examples/browser-to-server/tests/test.js
ddimaria: I like this suggestion ❤️   It's implemented.

examples/browser-to-server/tests/test.js
ddimaria: `"hello"` is now `'hello'` on line 73.

MarcoPolo (approved):

PR #109, opened 2023-04-03
fix: Fetch local fingerprint from SDP

ckousik:
Since firefox does not support `getFingerprints`, we fetch the local fingerprint from the PeerConnection's `localDescription` field by parsing the fingerprint line from the SDP.


src/sdp.ts
```diff
@@ -7,6 +7,10 @@ import type { HashCode, HashName } from 'multihashes'
 import { inappropriateMultiaddr, invalidArgument, invalidFingerprint, unsupportedHashAlgorithm } from './error.js'
 import { CERTHASH_CODE } from './transport.js'
 
+// import { detect } from 'detect-browser'
+
+// const browser = detect()
+
```
mxinden, line 13: Left over to be removed, right?

src/transport.ts
```diff
@@ -210,19 +210,12 @@ export class WebRTCTransport implements Transport {
       throw invalidArgument('no local certificate')
     }
 
-    const localCert = pc.getConfiguration().certificates?.at(0)
-
-    if (localCert === undefined || localCert.getFingerprints().length === 0) {
-      throw invalidArgument('no fingerprint on local certificate')
-    }
```
mxinden, line 217: Not sure how this is handled in the js-libp2p codebase in general. Should we still use `localCert.getFingerprints` in case the API is available (e.g. on Chrome), or should we always use the hack through the SDP?

mxinden (commented):
Thanks for tackling this.

mxinden:
@achingbrain @MarcoPolo any objections to merging here?

src/transport.ts
MarcoPolo: I think I would prefer trying to use the best option and falling back on the hack. But not blocking this.

MarcoPolo (commented):
Can we have firefox run this test in CI?

src/transport.ts
SgtPooki: I agree with Marco here, but it would be good to know (debug logs) which option was used if we're going to have different paths.

src/sdp.ts
```diff
@@ -15,6 +15,23 @@ const log = logger('libp2p:webrtc:sdp')
...
+const fingerprintRegex = /^a=fingerprint:(?:\w+-[0-9]+)\s(?<fingerprint>(:?[0-9a-fA-F]{2})+)$/gm
+function getFingerprintFromSdp (sdp: string): string | undefined {
+  const searchResult = fingerprintRegex.exec(sdp)
+  if (searchResult == null) {
+    return ''
+  }
+
+  return searchResult.groups?.fingerprint
+}
```
SgtPooki, line 34: Prefer `regex.match`.

Using `exec` can cause unexpected issues with the stored `lastIndex` being updated and then causing match failures in later `.exec` calls. 

> JavaScript [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) objects are stateful when they have the [global](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/global) or [sticky](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky) flags set (e.g. /foo/g or /foo/y). They store a [lastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex) from the previous match. Using this internally, exec() can be used to iterate over multiple matches in a string of text (with capture groups), as opposed to getting just the matching strings with [String.prototype.match()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match).

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#description

Also see https://stackoverflow.com/a/3811949/592760

SgtPooki (changes requested):

src/sdp.ts
```diff
@@ -15,6 +15,23 @@ const log = logger('libp2p:webrtc:sdp')
...
+  const localDescription = pc.localDescription
+  if (localDescription == null) {
+    return undefined
+  }
+  return getFingerprintFromSdp(localDescription.sdp)
+}
+
+const fingerprintRegex = /^a=fingerprint:(?:\w+-[0-9]+)\s(?<fingerprint>(:?[0-9a-fA-F]{2})+)$/gm
+function getFingerprintFromSdp (sdp: string): string | undefined {
+  const searchResult = fingerprintRegex.exec(sdp)
```
SgtPooki, line 28:
```suggestion
const fingerprintRegex = /^a=fingerprint:(?:\w+-[0-9]+)\s(?<fingerprint>(:?[0-9a-fA-F]{2})+)$/m
function getFingerprintFromSdp (sdp: string): string | undefined {
  const searchResult = sdp.match(fingerprintRegex)
```

src/sdp.ts
ckousik: Thanks @SgtPooki . I was not aware of this. I've incorporated the change and added a simple test.

src/sdp.ts
ckousik: Fixed

src/transport.ts
ckousik: @SgtPooki I've added the fallback to fetching from the SDP as requested.

p-shahi:
@MarcoPolo ping to re-review

ckousik:
@MarcoPolo I'm still working on firefox browser-to-server tests in CI. Could we move that work to a subsequent PR?

ckousik:
@MarcoPolo The browser-to-server tests is running on Firefox in CI.

src/sdp.ts
```diff
@@ -15,6 +15,42 @@ const log = logger('libp2p:webrtc:sdp')
...
+
+const fingerprintRegex = /^a=fingerprint:(?:\w+-[0-9]+)\s(?<fingerprint>(:?[0-9a-fA-F]{2})+)$/m
+export function getFingerprintFromSdp (sdp: string): string | undefined {
+  // const searchResult = fingerprintRegex.exec(sdp)
+  const searchResult = sdp.match(fingerprintRegex)
+  if (searchResult == null) {
+    return ''
+  }
+
+  return searchResult.groups?.fingerprint
```
SgtPooki, line 52: `'' == null // false`

Returning `''` on line 49 will cause the check on line 215 in src/transports.ts, `if (localFingerprint == null) {` to return false. We should return undefined here, or even better:

```suggestion
  const searchResult = sdp.match(fingerprintRegex)

  return searchResult?.groups?.fingerprint
```

test/sdp.spec.ts
```diff
@@ -53,6 +53,11 @@ describe('SDP', () => {
     ])
   })
 
+  it('extracts a fingerprint from sdp', () => {
```
SgtPooki, line 56: we should add a negative test here to validate the fallback behaviour

SgtPooki (changes requested):
i'm not an expert in the business logic of what is going on here, but i ran through the code and found a single potential failure that should be addressed.

src/sdp.ts
ckousik: Thank you!

test/sdp.spec.ts
ckousik: this just tests the regex behaviour. While there are no explicit tests for the fallback behaviour, it gets tested when the tests are run in Firefox as part of CI.

SgtPooki (commented):
Lgtm but i’ll defer to primary maintainers for final approval

p-shahi:
@MarcoPolo bump!

src/transport.ts
```diff
@@ -162,6 +163,18 @@ export class WebRTCDirectTransport implements Transport {
...
     // wait for peerconnection.onopen to fire, or for the datachannel to open
     const handshakeDataChannel = await dataChannelOpenPromise
+    const wrappedChannel = new WebRTCStream({ channel: handshakeDataChannel, stat: { direction: 'inbound', timeline: { open: 1 } } })
+    const wrappedDuplex = {
+      ...wrappedChannel,
+      sink: wrappedChannel.sink.bind(wrappedChannel),
+      source: {
+        [Symbol.asyncIterator]: async function * () {
+          for await (const list of wrappedChannel.source) {
+            yield list.subarray()
```
MarcoPolo: This may copy unnecessarily (I know this was here before).

MarcoPolo (approved):

examples/browser-to-server/package.json
```diff
@@ -7,7 +7,9 @@
     "start": "vite",
     "build": "vite build",
     "go-libp2p-server": "cd ../go-libp2p-server && go run ./main.go",
-    "test": "npm run build && playwright test tests"
+    "test": "npm run build && npm run test:chrome && npm run test:firefox",
+    "test:chrome": "playwright test tests",
+    "test:firefox": "npm run build && playwright test --browser firefox tests"
```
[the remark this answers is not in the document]
MarcoPolo: `aegir test` usually builds before testing. This is in line with that. We should probably build in the `test:chrome` as well.

Would be nice to have a test script that builds once then runs all the tests.

src/stream.ts
```diff
@@ -213,6 +213,7 @@ export class WebRTCStream implements Stream {
 
   constructor (opts: StreamInitOpts) {
     this.channel = opts.channel
+    this.channel.binaryType = 'arraybuffer'
```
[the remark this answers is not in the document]
MarcoPolo: It should be since the default is supposedly a blob: https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/binaryType. However I see the binaryType as arraybuffer when I create the channel. So I'm not sure why. I think we should keep this anyways.

MarcoPolo:
Thanks a lot @SgtPooki! Your review was very helpful :)

And thanks @wemeetagain for taking a look

MarcoPolo (approved):

PR #121, opened 2023-04-10
Fix multiaddr in webrtc connection

MarcoPolo:
The multiaddr in the webrtc connection would previously not include the webrtc component. This was confusing since if you wanted to see if a connection was over webrtc, checking the multiaddr alone would be wrong.

This sets the multiaddr on a webrtc connection to be `<boostrapMultiaddr>/webrtc/p2p/<peerid>`.

### DIFF
--- a/src/peer_transport/transport.ts
+++ b/src/peer_transport/transport.ts
@@ -4,6 +4,7 @@ import type { TransportManager, Upgrader } from '@libp2p/interface-transport'
 import { multiaddr, Multiaddr, protocols } from '@multiformats/multiaddr'
 import type { IncomingStreamData, Registrar } from '@libp2p/interface-registrar'
 import type { PeerId } from '@libp2p/interface-peer-id'
+import { peerIdFromString } from '@libp2p/peer-id'
 import { WebRTCMultiaddrConnection } from '../maconn.js'
 import type { Startable } from '@libp2p/interfaces/startable'
 import { WebRTCPeerListener } from './listener.js'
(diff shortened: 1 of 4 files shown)


src/peer_transport/transport.ts
```diff
@@ -75,15 +76,7 @@ export class WebRTCTransport implements Transport, Startable {
...
-   * and proceeds to upgrade to a webrtc connection.
-   * multiaddr of the form: <multiaddr>/webrtc/p2p/<destination-peer>
-   * For a circuit relay, this will be of the form
-   * <relay address>/p2p/<relay-peer>/p2p-circuit/webrtc/p2p/<destination-peer>
-  */
-  async dial (ma: Multiaddr, options: DialOptions): Promise<Connection> {
-    log.trace('dialing address: ', ma)
+  private splitAddr (ma: Multiaddr): { baseAddr: Multiaddr, peerId: PeerId } {
     const addrs = ma.toString().split(`${TRANSPORT}/`)
     if (addrs.length !== 2) {
```
maschad, line 81: not directly related but encountered in debugging, I think it's more intuitive to check if this is a `multiaddr` and then do the split.

```suggestion
if (!isMultiaddr(ma)) {
      throw new CodeError('invalid multiaddr', codes.ERR_INVALID_MULTIADDR)
}

const addrs = ma.toString().split(`${TRANSPORT}/`)
```

maschad (commented):
I made suggestion  based on the idea that should we decide to change how multiaddrs are formed in the future, I think this would be a more robust validation. Otherwise looks good to me.

src/peer_transport/transport.ts
MarcoPolo: This should be handled by the type system. We know this is a `Multiaddr` type in the param. When did you run into this not being a multiaddr?

The split and length check is to check if the `/webrtc/` part is in there. Really this should be interpreting the multiaddr from right to left, but the convention in JS has seemed to be to split on the string.

maschad: > This should be handled by the type system. We know this is a Multiaddr type in the param. When did you run into this not being a multiaddr? 

Yes you're correct, silly oversight on my part 😓 

> he split and length check is to check if the /webrtc/ part is in there. Really this should be interpreting the multiaddr from right to left, but the convention in JS has seemed to be to split on the string.

Correct I would say this is a bit more intuitive, as someone reading this code I can't understand immediately why a post-split array with a length that's not equal to 2, is an invalid multiaddr.  Thanks for clarifying.

maschad (approved):

PR #134, opened 2023-04-20
fix: added peer connection state listener to emit closed events

maschad:
The `RTCPeerConnection` state is not being monitored, as a result the `peer:Disconnect` event which is triggered by a `connectionEnd` event is not fired, resulting in closed connections not being properly propogated.

We can listen for disconnect events using the [RTCPeerConnection connectionstatechange event](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/connectionstatechange_event)

Closes: https://github.com/libp2p/js-libp2p-webrtc/issues/138

### DIFF
--- a/src/transport.ts
+++ b/src/transport.ts
@@ -184,6 +184,23 @@ export class WebRTCDirectTransport implements Transport {
       }
     }
 
+    const eventListeningName = isFirefox ? 'iceconnectionstatechange' : 'connectionstatechange'
+
+    peerConnection.addEventListener(eventListeningName, () => {
+      switch (peerConnection.connectionState) {
+        case 'failed':
+        case 'disconnected':
+        case 'closed':
+          maConn.close().catch((err) => {
+            log.error('error closing connection', err)
+          })
+          break
+
+        default:
+          break
+      }
+    })
+
     // Creating the connection before completion of the noise
     // handshake ensures that the stream opening callback is set up
     const maConn = new WebRTCMultiaddrConnection({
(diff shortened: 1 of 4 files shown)


src/transport.ts
```diff
@@ -189,10 +190,42 @@ export class WebRTCDirectTransport implements Transport {
       peerConnection,
       remoteAddr: ma,
       timeline: {
-        open: (new Date()).getTime()
+        open: Date.now()
       }
     })
 
+    if (isFirefox) {
+      peerConnection.addEventListener('oniceconnectionstatechange', () => {
```
ddimaria, line 198: You could remove some duplication here:
```
const stateChangeEvent = isFirefox ? 'oniceconnectionstatechange' : 'onconnectionstatechange';

peerConnection.addEventListener(stateChangeEvent, () => {
....
}
```

src/transport.ts
```diff
@@ -183,13 +184,30 @@ export class WebRTCDirectTransport implements Transport {
       }
     }
 
+    const eventListeningName = isFirefox ? 'iceconnectionstatechange' : 'connectionstatechange'
```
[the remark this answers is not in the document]
maschad: Good point, well `iceconnectionstatechange` event has been supported since [Firefox 24](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/iceconnectionstatechange_event) which is about a decade old, and `connectionstatechange` has been supported for [at least two years](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/connectionstatechange_event) in all other browsers, so I would say this is low risk.

src/transport.ts
```diff
@@ -183,13 +184,31 @@ export class WebRTCDirectTransport implements Transport {
...
+
+    peerConnection.addEventListener(eventListeningName, () => {
+      switch (peerConnection.connectionState) {
+        case 'failed':
+        case 'disconnected':
+        case 'closed':
+          maConn.close().catch((err) => {
+            log.error('error closing connection', err)
+          }).finally(() => {
+            peerConnection.removeEventListener(eventListeningName, () => {})
```
MarcoPolo, line 197: This doesn't remove this function, right? You're passing a new function here that doesn't reference any existing listener. You have to name the callback handler to remove it.

MarcoPolo: Alternatively you can use the signal in the options to addEventListener. e.g.:

```js
const controller = new AbortController();
const signal = controller.signal;
element.addEventListener('eventname', (ev) => {
    // do stuff...
}, { signal });
// To remove the event listener:
controller.abort();
```

> Event listeners can also be removed by passing an [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) to an [addEventListener()](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) and then later calling [abort()](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) on the controller owning the signal.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

MarcoPolo (approved):

Issue #144, opened 2023-04-27
Fails to send with high bitrate (or large messages)


marcus-pousette:
I am running into issues when using WebRTC instead of Websockets from a browser. 

Using WebRTC transport seem to fail if the bitrate is too high, and the failure is kind of "silent", which makes it hard to detect. 

I am able to reproduce in the browser-to-browser example, 

1). 

Set 
```
streamMuxers: [mplex({ maxMsgSize: 15 * 1e3 })], 
```
Based on the [message size limitation](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_data_channels#understanding_message_size_limits)

2). 
Instead of sending a message of text, send 1mb of data ( ```sender.push(new Uint8array(1e6))```)
Log the results on the other side. 

Sending data works when the message is small (isch). < 200 kb. Fails otherwise. I am running a Chrome browser 

In my own application I can run the exact same setup with Websockets and a relay, with mplex message size limitation above and it works as expected. But it does not work with WebRTC transport when the bitrate becomes too large (or the message sizes).

marcus-pousette:
Okey, so this transport already has a [muxer](https://github.com/libp2p/js-libp2p-webrtc/blob/main/src/muxer.ts). Looks like it does not support chunking like mplex, with 'maxMsgSize'. I guess it is still possible to do hear since the data messages are length prefixed.

maschad:
Currently, it's not practical to use `RTCDataChannel` for messages larger than 64 KiB (16 KiB if you want to support cross-browser exchange of data - [That is why in the spec states that encoded messages including their length prefix **MUST NOT** exceed 16kiB to support all major browsers](https://github.com/libp2p/specs/tree/master/webrtc#multiplexing).) 

The problem arises from the fact that SCTP—the protocol used for sending and receiving data on an RTCDataChannel—was originally designed for use as a signaling protocol. There is a [proposal](https://datatracker.ietf.org/doc/html/rfc8260) which will make it possible to send messages with essentially no size limitations, but for now perhaps we should just set that max limit to 16Kb to avoid silent failures for consumers.

maschad:
[...]

PR #147, opened 2023-04-28
feat: restrict message sizes to 16kb 

marcus-pousette:
Solves #144 

Currently, not only do one need to push messages in parts to the datachannel, but also one needs to throttle the send function so that the datachannel buffer does not exceed ```x``` bytes. If x > 256kb the datachannel will close if you are using Chrome

The unit tests added are not testing what happens when the bufferedAmount exceeds maxMsgSize even though there is functionality for that. 
 
I tested this solution in the browser-to-browser example also by sending large messages (1mb) instead of text. 

Ideally, there would be a default max msg size limit would already set so that new users are not running into this when using this transport. The default behaviour is very unpredictable as it fails on different conditions depending on what browser combination used, message sizes and messaging frequency.


maschad (changes requested):
Thanks for this PR,  this is great. Following up from my comment https://github.com/libp2p/js-libp2p-webrtc/issues/144#issuecomment-1540979091 could you refactor this to exclude the optional `maxMsgSize` and instead include the changes that ensure the underlying buffer doesn't exceed the max size of 16kb.

marcus-pousette:
I have now rebased and hardcoded 16kb limit

mxinden:
- Restricting the size of the message passed to [`RTCDataChannel.send`](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/send) to the minimum support 16kib sounds good to me.
  - See also corresponding limit in rust-libp2p https://github.com/libp2p/rust-libp2p/blob/1742bffcea4691f013e966393ae0d7fe95f8834c/transports/webrtc/src/tokio/substream.rs#L45-L50
- I am in favor of splitting messages according to the above.
- Do I understand correctly that this pull request effectively restricts **both** the message size and the `RTCDataChannel` `bufferedAmount` to never exceed 16kib?
- If I am not mistaken, given that the `MessageProcessor` simply buffers bytes exceeding the `bufferedAmount` limit in `sendMessageQueue`, there is no backpressure. But I guess, without being familiar with most of the APIs in `js-libp2p` there is no backpressure on the read or write path in the first place, correct?

marcus-pousette:
> * Do I understand correctly that this pull request effectively restricts **both** the message size and the `RTCDataChannel` `bufferedAmount` to never exceed 16kib?

Yes. I am not fully confident what the limits should be, and it is quite hard to tests, frankly. I made this decision reading various threads about sending large files over WebRTC and what has been done to circumvent issues.

Max buffered amount should maybe be 16 Mbytes  instead
https://viblast.com/blog/2015/2/25/webrtc-bufferedamount/
https://groups.google.com/g/discuss-webrtc/c/EjevtDTsxuE/m/eowQM2wNAwAJ

Some other discussion 
https://github.com/feross/simple-peer/issues/393


> * If I am not mistaken, given that the `MessageProcessor` simply buffers bytes exceeding the `bufferedAmount` limit in `sendMessageQueue`, there is no backpressure. 

Yes

> But I guess, without being familiar with most of the APIs in `js-libp2p` there is no backpressure on the read or write path in the first place, correct?

Not sure

achingbrain (changes requested):
Instead of buffering, you can slow down the ingestion of new data from the source in the `_sink` method using [p-event](https://www.npmjs.com/package/p-event):

```js
for await (const buf of merge(closeWrite, src)) {
  if (channel.bufferedAmount > maxMsgSize) {  // this should really be maxBufferedAmountSize
    await pEvent(this.channel, 'bufferedamountlow', {
      timeout: // a sensible timeout
    })
  }
  
  // now write message as normal
  //... chunk using maxMsgSize
```

There's no need to buffer anything then as backpressure is implicitly applied since we stop pulling from the incoming source.

When sending the message you can then limit the max size of individual message by doing chunking in a [similar way to mplex](https://github.com/libp2p/js-libp2p-mplex/blob/master/src/stream.ts#L177-L185).

marcus-pousette:
> Instead of buffering, you can slow down the ingestion of new data from the source in the `_sink` method using [p-event](https://www.npmjs.com/package/p-event):

Thats a great idea. 

I made the changes now. Not done tests for the max buffered amount, not sure how to do it cleanly.

src/stream.ts
```diff
@@ -151,6 +152,15 @@ class StreamState {
...
 }
 
+// Max message size that can be sent to the DataChannel
+const MAX_MESSAGE_SIZE = 16 * 1024
+
+// How much can be buffered to the DataChannel at once
+const MAX_BUFFERED_AMOUNT = 16 * 1024 * 1024
+
+// How long time we wait for the 'bufferedamountlow' event to be emitted
+const BUFFERED_AMOUNT_LOW_TIMEOUT = 30 * 1000
```
achingbrain, line 162: These are fine as defaults (I think) but the the user needs to be able to override them.

When creating the transport they should be able to pass these in as options:

```js
const node = await createLibp2p({
  transports: [
    webRTC({
      maxDataChannelMessageSize: number,
      maxDataChannelBufferedAmount: number,
      dataChannelBufferAmountLowEventTimeout: number
    })
     //.. other transports
  ]
  //.. other config
})
```

They should be added to the [WebRTCTransportInit](https://github.com/libp2p/js-libp2p-webrtc/blob/main/src/private-to-private/transport.ts#L24-L26) interface with some appropriate JSDoc comments explaining what they do so users get nice IntelliSense-style help prompts when using them.

test/stream.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    send: cb.send,
+    close: () => {}
+  }
+  return new underTest.WebRTCStream({ channel: datachannel as RTCDataChannel, stat: underTest.defaultStat('outbound') })
+}
+
+const MAX_MESSAGE_SIZE = 16 * 1024
+
+describe('Max message size', () => {
+  it(`sends messages smaller or equal to ${MAX_MESSAGE_SIZE} bytes in one`, async () => {
```
achingbrain, line 20: Thank you for adding tests.

Could you please add one more that tests that when the event timeout is reached an appropriate error is thrown and that the stream is closed correctly.

achingbrain (changes requested):
Nearly there - just a couple of extra bits to do, comments are inline.

src/stream.ts
marcus-pousette: Add added the args as 

```typescript
 webRTC({
      constraints: {
          maxDataChannelMessageSize: number,
          maxDataChannelBufferedAmount: number,
          dataChannelBufferAmountLowEventTimeout: number
      }
    })
```

else moving the options around not that nice... 

More minimalistic woiuld be
```typescript
 webRTC({
      ...other,
      datachannel: {
          maxMessageSize: number,
          maxBufferedAmount: number,
          timeout: number
      }
    })
```

?


Also it kind of also make sense to add a test to see whether the arguments are propagating through correctly, both when you initiate a connection, and also when you are on the receiving end.

achingbrain: If having everything as a top-level option is beyond the pale, I think this is preferable:

```js
 webRTC({
      ...other,
      dataChannel: {
          maxMessageSize: number,
          maxBufferedAmount: number,
          timeout: number
      }
    })
```

This way it's clear the limits apply to the data channel, but `timeout` needs something to disambiguate it - my intuition would be that it's a stream inactivity timeout but that's not the case here.

test/stream.spec.ts
achingbrain: Agreed, I added an e2e test here but it doesn't appear to work 😱 - https://github.com/libp2p/js-libp2p-webrtc/pull/165

test/stream.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    try {
+      await webrtcStream.sink(p)
+      throw new Error('Expected to fail')
+    } catch (error: any) {
+      expect(error.message).eq('Timed out waiting for DataChannel buffer to clear')
+      const t1 = Number(new Date())
+      expect(t1 - t0).greaterThan(timeout)
+      expect(t1 - t0).lessThan(timeout + 1000) // Some upper bound
+      expect(closed).true()
+    }
```
achingbrain, line 110:
```suggestion
    await expect(webrtcStream.sink(p)).to.eventually.be.rejected
      .with.property('message', 'Timed out waiting for DataChannel buffer to clear')

    const t1 = Number(new Date())
    expect(t1 - t0).greaterThan(timeout)
    expect(t1 - t0).lessThan(timeout + 1000) // Some upper bound
    expect(closed).true()
```

src/stream.ts
marcus-pousette: Soo this is a ```timeout``` for unresponsiveness. But generally ```timeout``` in connections means that no party has sent any message for a while, so we close the connection. So perhaps this makes sense
```typescript
 webRTC({
      ...other,
      
      dataChannel: {
          maxMessageSize: number,
          maxBufferedAmount: number,
          bufferTimeout: number,
          keepAliveTimeout: number, // not a feature now, but in theory
      }
    })

```

test/stream.spec.ts
marcus-pousette: I did tests manually now, logging the configurations on both ends when runninng the browser-to-browser example and saw that the configurations are propagating through

src/private-to-public/transport.ts
```diff
@@ -231,7 +231,7 @@ export class WebRTCDirectTransport implements Transport {
     // Track opened peer connection
     this.metrics?.dialerEvents.increment({ peer_connection: true })
 
-    const muxerFactory = new DataChannelMuxerFactory(peerConnection, this.metrics?.dialerEvents)
+    const muxerFactory = new DataChannelMuxerFactory({ peerConnection, metrics: this.metrics?.dialerEvents })
```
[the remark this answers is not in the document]
marcus-pousette: I just never added it here, as I initially only was interested in fixing the browser-to-browser case. But I fixed it now.

test/stream.spec.ts
marcus-pousette: I also addded it to ```wrappedChannel``` now, I am not sure the implications of this, if it is only for the noise handshake?

src/private-to-public/transport.ts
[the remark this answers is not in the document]
marcus-pousette: I also addded it to ```wrappedChannel``` now, I am not sure the implications of this, if it is only for the noise handshake?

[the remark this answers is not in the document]
achingbrain: It shouldn't hurt - the noise handshake should be well under the 16kb limit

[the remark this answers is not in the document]
marcus-pousette: Good!

test/stream.spec.ts
marcus-pousette: Would my manual test be fine for now before #165 is fixed? 

Perhaps we can add proper configuration testing later, also add bound checking/tests for configuration, so you don't create a configuration with unfeasible constraints

achingbrain: I think it's fine - we can more comprehensive tests after #165

test/stream.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      stat: underTest.defaultStat('outbound')
+    })
+
+    p.push(data)
+    p.end()
+    await webrtcStream.sink(p)
+    expect(sent).to.deep.equals([messageLengthEncoded])
+  })
+
+  it(`sends messages greather ${MAX_MESSAGE_SIZE} bytes in parts`, async () => {
```
achingbrain, line 48: Typo

```suggestion
  it(`sends messages greater than ${MAX_MESSAGE_SIZE} bytes in parts`, async () => {
```

PR #160, opened 2023-05-10
deps!: update all libp2p deps for compat with libp2p@0.45.x

achingbrain:
- Use event bus to populate listened-on addresses
- Derive the connection remote address from the SDP offer so the remote address is a webrtc address and not related to the underlying transport that was used to do the SDP handshake
- Simplify example tests with `test-ipfs-example` - no need to install playwright separately
- Update browser to browser example test to test without the relay
- Set connection close time when RTCPeerConnection closes
- Updates all streaming deps
- Removes unused peer-store dependency
- Updates project config and fixes badges
- Updates ci files (will be automated by https://github.com/protocol/.github/pull/519 )


maschad (approved):

achingbrain:
Just tidying up the examples and fixing some weirdness, push incoming..

maschad (approved):

src/maconn.ts
```diff
@@ -64,6 +64,12 @@ export class WebRTCMultiaddrConnection implements MultiaddrConnection {
     this.remoteAddr = init.remoteAddr
     this.timeline = init.timeline
     this.peerConnection = init.peerConnection
+
+    this.peerConnection.onconnectionstatechange = () => {
```
maschad, line 68: Not sure if this is necessary, we have [an event listener](https://github.com/libp2p/js-libp2p-webrtc/blob/4f1840e04da926e49ae24f5f7639780a1aef4155/src/transport.ts#L216-L217) that will close these connections in the transport as well as in the [peer transport](https://github.com/libp2p/js-libp2p-webrtc/blob/4f1840e04da926e49ae24f5f7639780a1aef4155/src/peer_transport/handler.ts#L43)

src/peer_transport/handler.ts
```diff
@@ -133,5 +135,21 @@ export async function initiateConnection ({ rtcConfiguration, signal, stream: ra
...
+  // 'a=candidate:1746876089 1 udp 2113937151 0614fbad-b...ocal 54882 typ host generation 0 network-cost 999'
+  const candidateLine = sdp.split('\r\n').filter(line => line.startsWith('a=candidate')).pop()
+  const candidateParts = candidateLine?.split(' ')
+
+  if (candidateLine == null || candidateParts == null || candidateParts.length < 5) {
+    log('could not parse remote address from', candidateLine)
+    return '/webrtc'
+  }
+
+  return `/dnsaddr/${candidateParts[4]}/${candidateParts[2]}/${candidateParts[3]}/webrtc`
```
maschad, line 154: Why are we returning the `remoteAddress` now? Would this be a breaking change?

maschad (changes requested):
Are some of the [other changes](https://github.com/libp2p/js-libp2p-webrtc/blob/4f1840e04da926e49ae24f5f7639780a1aef4155/src/peer_transport/handler.ts#L43) introduced required for compatibility? If not maybe we should move them to a separate PR

src/peer_transport/handler.ts
achingbrain: This is fix.  Previously we returned the relayed address, so it looked like the remote webrtc connection had an address like `/ip4/123.123.132.123/p2p/QmRelay/webrtc/p2p/QmPeer` when in fact only the SDP handshake occurred over that address, the actual connection was over webrtc.

This can be validated by connecting two browser peers via a relay then stopping the relay - the connection is not tunnelled via the relay so it shouldn't be in the remote addr of the connection.

src/maconn.ts
achingbrain: We need something else - that event listener only handles outbound connections.  We need to handle inbound connections too.

maschad (approved):

PR #165, opened 2023-05-15
fix: use abstract stream class from muxer interface module

### LINKED ISSUE #164: Streams are never closed
The following simple echo stream code does not work:

```js
import all from 'it-all'

const stream = await connection.newStream(echo)
const input = new Array(5).fill(0).map(() => new Uint8Array(10))

const output = await pipe(
  input,
  stream,
  all
)
console.info('all done') // <- never reached
```

The reason is the internal `_sink` method merges the source stream with a "close write iterable" that waits for the `closeWritePromise` deferred promise to resolve.  This is done in the [`finally` block](https://github.com/libp2p/js-libp2p-webrtc/blob/main/src/stream.ts#L312) of the `sink` method.

This block is never reached because the [`this._sink` call](https://github.com/libp2p/js-libp2p-webrtc/blob/main/src/stream.ts#L301) doesn't return - it's waiting for the `closeWritePromise` deferred promise to resolve which creates a deadlock.

achingbrain:
Refactors the stream class in this module to use the abstract superclass from `@libp2p/interface-stream-muxer` as it handles all the various scenarios for closing streams which this module does not.

Fixes #164


achingbrain:
Refs: https://github.com/libp2p/js-libp2p-webrtc/issues/164

maschad (approved):
LGTM

PR #168, opened 2023-05-18
chore: more consistent imports

vmx:
Put imports from the same file into a single line.


achingbrain:
I think this should be handled by [linting rules](https://github.com/ipfs/eslint-config-ipfs/) otherwise this is a never-ending source of PRs that neither fix bugs or add features and consequent release churn.

vmx:
Cool, I close it then and wait for better lints.

PR #170, opened 2023-05-22
fix: make it work as GiHub dependency when using Vite

vmx:
When using `@libp2p/webrtc` directly with pointing to a GitHub repository, it won't build with Vite, due to the defined `exports` in the package.json. One way to make it work is adding the build step as `prepare` script.

Steps to reproduce the issue:

    npm create vite@latest -- --template vanilla webrtcasgithubdep
    cd webrtcasgithubdep
    npm install https://github.com/libp2p/js-libp2p-webrtc
    echo "import { webRTC } from '@libp2p/webrtc'" > main.js
    npm run build

Errors with:

    [commonjs--resolver] Failed to resolve entry for package "@libp2p/webrtc". The package may have incorrect main/module/exports specified in its package.json.
error during build

With the prepare script it builds without errors.

---

There are probably better fixes and it might even be something to do on the Aegir level. I thought I open a PR as this is a good way to also try it out. Just follow those steps and then run

    npm install https://github.com/vmx/js-libp2p-webrtc#add-prepare-script
    npm run build

and you can see the it builds successfully.

Thanks @Gozala for pointing me to this solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
   "scripts": {
     "generate": "protons src/private-to-private/pb/message.proto src/pb/message.proto",
     "build": "aegir build",
+    "prepare": "npm run build",
     "test": "aegir test -t browser",
     "test:chrome": "aegir test -t browser --cov",
     "test:firefox": "aegir test -t browser --browser firefox",


maschad (changes requested):
Have you tried loading the module as [a Browser script tag](https://github.com/libp2p/js-libp2p-webrtc#browser-script-tag) ?  The use case presented here is a bit specific whereas the suggested ways to utilize the js-libp2p ecosystem is through npm installation or script tags.

vmx:
I would argue that many people are using some sort of bundlers for their work in browsers (it's hardly possible without these days). Also the [js-libp2p browser example is using Vite](https://github.com/libp2p/js-libp2p/tree/99dcc6f148a12fc7efd4c44bbf313ecbd9d68568/examples/libp2p-in-the-browser).

maschad:
> I would argue that many people are using some sort of bundlers for their work in browsers (it's hardly possible without these days). Also the [js-libp2p browser example is using Vite](https://github.com/libp2p/js-libp2p/tree/99dcc6f148a12fc7efd4c44bbf313ecbd9d68568/examples/libp2p-in-the-browser).

If that's the case then why not install the dep from npm as opposed to Github? That's not really a common use case.

maschad:
> I agree it's not common for end-users. Though I use that a lot during development.

I can appreciate that but I think in the interest of making the project standardized like our other libraries it isn't applicable. This change is more specific to your development workflow and as such I would suggest that they remain in your forked repo as opposed to the main repo.

PR #176, opened 2023-06-02
test: actually run firefox tests on firefox

achingbrain:
I noticed the `--browser` was not being passed to playwright test as a forward arg, so ensure that it's actually being run on the browser we think it is.


MarcoPolo:
I was wondering why firefox was passing here, but not in the interop test!

PR #175, opened 2023-06-02
test: add a test for large transfers

achingbrain:
Tests that we can transfer messages well over the buffer limit.

PR #177, opened 2023-06-02
test: run tests on webkit

achingbrain:
Adds a `test:webkit` script so we can run tests on webkit in CI.


test/basics.spec.ts
```diff
@@ -58,15 +58,16 @@ describe('basics', () => {
     remoteNode = await createNode()
   })
 
-  afterEach(async () => {
-    if (localNode != null) {
-      await localNode.stop()
-    }
+  // This is currently causing issues on macOS runs
```
maschad, line 61: There seems to be a non-deterministic bug during the teardown of our relay tests which results in an error being thrown on webkit tests, specifically when running on my local (macOS) where the `RTCDataChannel` is unable to [send the encoded `STOP_SENDING` flag](https://github.com/libp2p/js-libp2p-webrtc/blob/test/run-tests-on-webkit/src/stream.ts#L190).  

``` bash
Error sending binary data through RTCDataChannel.
```
See image for more details.
image

MarcoPolo: Are you sure we're setting the binaryType correctly everytime we create a datachannel? I think we might be missing these:

```
diff --git a/src/private-to-private/handler.ts b/src/private-to-private/handler.ts
index 176031b..793477e 100644
--- a/src/private-to-private/handler.ts
+++ b/src/private-to-private/handler.ts
@@ -104,6 +104,7 @@ export async function initiateConnection ({ rtcConfiguration, dataChannelOptions
   // to collect candidates. The label is not relevant to connection initiation
   // but can be useful for debugging
   const channel = pc.createDataChannel('init')
+  channel.binaryType = 'arraybuffer'
   // setup callback to write ICE candidates to the remote
   // peer
   pc.onicecandidate = ({ candidate }) => {
diff --git a/src/private-to-public/transport.ts b/src/private-to-public/transport.ts
index c23b715..ac4fdd7 100644
--- a/src/private-to-public/transport.ts
+++ b/src/private-to-public/transport.ts
@@ -139,6 +139,7 @@ export class WebRTCDirectTransport implements Transport {
     // the peer.
     const dataChannelOpenPromise = new Promise<RTCDataChannel>((resolve, reject) => {
       const handshakeDataChannel = peerConnection.createDataChannel('', { negotiated: true, id: 0 })
+      handshakeDataChannel.binaryType = 'arraybuffer'
       const handshakeTimeout = setTimeout(() => {
         const error = `Data channel was never opened: state: ${handshakeDataChannel.readyState}`
         log.error(error)

```

maschad: I think so as [all streams are initialized as such](https://github.com/libp2p/js-libp2p-webrtc/blob/test/run-tests-on-webkit/src/stream.ts#L66) and even if it was set to `blob` ([the default](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/binaryType)) then we would have seen errors being thrown in [other tests](https://github.com/libp2p/js-libp2p-webrtc/blob/test/run-tests-on-webkit/test/stream.browser.spec.ts#L12) but I did a sanity check and added those changes as well and the result was the same. As I had mentioned the errors are thrown **after** the teardown of the tests - i.e. the test actually passes.

maschad: I do notice that also [the setup was moved on this PR as well](https://github.com/libp2p/js-libp2p-webrtc/pull/175/files#diff-82800680dd98113b2d5f557f41fbfca2a686b7dc84fe106e432d86367a8d15a5R88-R96) I am not sure if that has any relation.

maschad:
So it seems there [isn't parity across the Linux Webkit implementations on playwright](https://github.com/microsoft/playwright/issues/4017) which was causing these tests to fail on a linux vm for webkit. There is currently [an open issue to add support for better testing WebRTC applications with Webkit](https://github.com/microsoft/playwright/issues/2973) but in the interim I suggest we run the webkit tests on a _macOS_ instance until that's resolved.

maschad:
> So it seems there [isn't parity across the Linux Webkit implementations on playwright](https://github.com/microsoft/playwright/issues/4017) which was causing these tests to fail on a linux vm for webkit. There is currently [an open issue to add support for better testing WebRTC applications with Webkit](https://github.com/microsoft/playwright/issues/2973) but in the interim I suggest we run the webkit tests on a [_macOS_ instance](https://github.com/libp2p/js-libp2p-webrtc/pull/177/commits/747b5855f6ceadfcdcf1abde965de94f6ef5a4e5) until that's resolved.

Update: So I did some more research/testing and found that WebkitGTK (the full-featured port of the WebKit rendering engine on Linux-based OS's) does not _yet_ support the [PeerConnection API](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection). Igalia is currently working on [the implementation](https://base-art.net/Articles/webrtc-in-webkitgtk-and-wpe-status-updates-part-i/) but this is obviously a blocker for Playwright supporting this interface. 

Thanks to @MarcoPolo for pointing me in this direction.

maschad (approved):

PR #178, opened 2023-06-03
docs: update README.md example

JoseKP:
`@chainsafe/libp2p-noise` package doesn't return `Noise` class anymore, instead is using `noice` function

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,11 @@ Loading this module through a script tag will make it's exports available as `Li
 
 ```js
 import { createLibp2p } from 'libp2p'
-import { noise } from "@chainsafe/libp2p-noise";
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ Loading this module through a script tag will make it's exports available as `Li
 
 ```js
 import { createLibp2p } from 'libp2p'
-import { Noise } from '@chainsafe/libp2p-noise'
+import { noise } from "@chainsafe/libp2p-noise";
```
maschad, line 47: Please use single quotes and remove the semi-colons otherwise our linter will fail

maschad (changes requested):
Looks good just a comment

maschad (approved):
Thanks again!

